Issue #350, opened 2013-09-21
Implement SVD++


ghost:
It'd be nice to have an implementation of Koren's SVD++ recommender.

Paper: http://dx.doi.org/10.1145/1401890.1401944

**Note**: This issue has been automatically migrated from Bitbucket
Created by @elehack on 2013-09-12T13:56:42.803781+00:00

PR #437, opened 2013-11-14
implementation of svd++ (#350)

qiankun925:
- add implementation of svd++
- train the latent features one by one
- TrainingEstimator in funksvd is changed to be public
- for the output model, users' features are added up with the normalized item implicit features


mdekstrand:
We are working on a new version of SVD++ now.

PR #617, opened 2014-08-11
Fix the crossfold by changing write files logic.

fxu12:
When crossing fold the data, it first divides users into different groups. The number of groups is given by configuration in eval.grovy. When we try to put a user's rating history into test set or train set,  we first get the fold number of which group this user belongs to and then get the cut off number that separates the user's ratings into train or test data.
But the original version of lenskit-crossfold iterates all the groups when writing a user's rating history, and if a user's group does not correspond to the file number which going to be written, it just writes all this user's rating history into the training set of that file, which is done in the else part we took off from the commitment.
The else part does not make sense to us.

If we did not understand wrong, is that the supposed behavior of cross-fold? Thanks!


elehack:
Just a heads-up, #656 fixes the defaults problem.

Issue #630, opened 2014-10-23
Order of data set columns not consistent across JVMs


elehack:
I'm seeing odd behavior running the evaluator across multiple JVMs.

On Java 6 and 7, the Partition column comes before DataSet. On Java 8, DataSet comes first (which is the better order, IMO).

This should be consistent.

PR #655, opened 2015-01-12
Used linked hash map in TTDataSet attributes

elehack:
Use a linked hash map to maintain the order of data set attributes in `GenericTTDataSet`. Addresses #630.


elehack:
No review needed, will merge when tests pass.

Issue #621, opened 2014-09-22
Simple evaluator and eval script have different crossfold defaults


elehack:
According to #617, the simple evaluator and eval scripts (via `CrossfoldTask`) have different defaults for the crossfolder's configuration.

This need to be fixed — the defaults should be the same across the board.

elehack:
This was completed by #656.

PR #656, opened 2015-01-13
Change simple evaluator default crossfold

elehack:
Address #621 by changing the simple evaluator defaults to use the crossfold task defaults.

- [x] Review
- [ ] Merge
- [ ] Update release notes to document incompatible change


elehack:
Not sure if we need doc changes on this, other than deleting the old default. Is it implied that an unspecified default uses the defaults in `CrossfoldTask`?

elehack:
The reason I care about not documenting is that the more places we document some magic number, the more places can go out of date. I will resolve by pointing directly to the `CrossfoldTask` docs.

Issue #563, opened 2014-04-10
Write a Known rating rating predictor


kluver:
Currently there is no way to easily make lenskit return known ratings when asked to predict an item with a known rating.

An easy solution would be a custom RatingPredictor that takes a fallback scorer or predictor, but returns ratings from the DAO if any are available.

Some further notes:

- should be in the `lenskit-predict` module
- should be called `KnownRatingRatingPredictor`
- will depend on a `UserEventDAO`, get the user's rating events from the DAO
- will score using the user's existing ratings

elehack:
I've seen initial code for this, it looks like it will land for 2.2.

PR #657, opened 2015-01-15
Known rating rating predictor 

tajinder-txstate:
Ticket #563

### DIFF
--- a/lenskit-predict/src/main/java/org/grouplens/lenskit/predict/KnownRatingRatingPredictor.java
+++ b/lenskit-predict/src/main/java/org/grouplens/lenskit/predict/KnownRatingRatingPredictor.java
@@ -1,3 +1,23 @@
+/*
+ * LensKit, an open source recommender systems toolkit.
+ * Copyright 2010-2014 LensKit Contributors.  See CONTRIBUTORS.md.
+ * Work on LensKit has been funded by the National Science Foundation under
+ * grants IIS 05-34939, 08-08692, 08-12148, and 10-17697.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
 package org.grouplens.lenskit.predict;
 
 import it.unimi.dsi.fastutil.longs.LongSortedSet;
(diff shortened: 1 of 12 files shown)


elehack:
The CI builds failed, with this error:

> License violations were found: /home/travis/build/lenskit/lenskit/lenskit-predict/src/main/java/org/grouplens/lenskit/predict/KnownRatingRatingPredictor.java

That means that a new file is missing its license header. You can identify these problems in advance by running the `check` Gradle task.

You can add the missing license header(s) by running the `licenseFormat` Gradle task. Then commit the changes and push them back to your `feature/563-known-predictor` branch, and the pull request will be automatically updated.

lenskit-predict/src/main/java/org/grouplens/lenskit/predict/KnownRatingRatingPredictor.java
```diff
@@ -0,0 +1,43 @@
...
+import org.grouplens.lenskit.data.event.Rating;
+import org.grouplens.lenskit.vectors.MutableSparseVector;
+
+import javax.annotation.Nonnull;
+import javax.inject.Inject;
+import java.util.List;
+
+/**
+ * Created by Tajinder on 1/6/2015.
+ * 'KnownRatingRatingPredictor' predicts the known ratings of each item and the user given in the predictions
```
elehack: Let's rewrite this comment. A couple of general principals:

The first sentence will be a brief summary. It should *not* include the class name, since the class name will always be displayed with it.

So something like this:

```
/**
 * Use a user's existing ratings as predictions.  This rating predictor 
 * returns the user's rating as a prediction for all items the user has rated, 
 * and returns no prediction for items the user has rated. It is useful in
 * conjunction with a {@link FallbackItemScorer} to return a user's rating
 * when it is known.
 *
 * @see RatingPredictorItemScorer
 * @since 2.2
 */

File: lenskit-predict/src/main/java/org/grouplens/lenskit/predict/KnownRatingRatingPredictor.java [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+ */
+public class KnownRatingRatingPredictor extends AbstractRatingPredictor {
+    private final UserEventDAO userEvent;
```
elehack: We should consider a clearer name for this variable. Perhaps `userEventDAO`, or just `dao` since there is only one DAO in use.

lenskit-predict/src/main/java/org/grouplens/lenskit/predict/KnownRatingRatingPredictor.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Inject
+    public KnownRatingRatingPredictor(UserEventDAO user) {
```
elehack: Again, naming: a user event DAO is not a 'user'. I often use 'uedao' for a short parameter name in these kinds of constructors.

lenskit-predict/src/main/java/org/grouplens/lenskit/predict/KnownRatingRatingPredictor.java
```diff
@@ -0,0 +1,43 @@
...
+        this.userEvent = user;
+    }
+
+    @Override
+    public void predict(long user, @Nonnull MutableSparseVector predictions) {
+        LongSortedSet tempPredictions = predictions.keyDomain();
```
elehack: This variable does not hold a temporary copy of predictions; instead, it holds the desired items. We can just call it `items`.

elehack:
Good work. The logic is solid and the test cases are good. I have put some comments in about the JavaDoc and about variable names.

There is one additional important case that we haven't tested for yet: handling an unrate. This can be done by having a user who has some ratings, and then an unrate for one of the rated items; you will need to use `RatingBuilder` to build the unrate event. These events will also need timestamps, and the unrate will need a timestamp after the rating's timestamp. Then we test that it does not produce a prediction.

lenskit-predict/src/main/java/org/grouplens/lenskit/predict/KnownRatingRatingPredictor.java
```diff
@@ -0,0 +1,70 @@
...
+public class KnownRatingRatingPredictor extends AbstractRatingPredictor {
+    private final UserEventDAO dao;
+
+    @Inject
+    public KnownRatingRatingPredictor(UserEventDAO uedao) {
+        this.dao = uedao;
+    }
+
+    @Override
+    public void predict(long user, @Nonnull MutableSparseVector predictions) {
```
elehack: Looks good!

lenskit-predict/src/test/java/org/grouplens/lenskit/predict/KnownRatingRatingPredictorTest.java
```diff
@@ -0,0 +1,129 @@
...
+                .setItemId(840)
+                .setRating(3.5)
+                .clearRating()
+                .build();
+        rs.add(r);
+
+        KnownRatingRatingPredictor KnownPredict = new KnownRatingRatingPredictor(userDAO);
+        MutableSparseVector predictItems = MutableSparseVector.create(840);
+        KnownPredict.predict(420, predictItems);
+        assertThat(predictItems.size(), equalTo(0));
```
elehack: Good start. But this test doesn't quite test the behavior we're looking for; we need an unrate event *that undoes a previous rating*. For this, we need the ratings to actually have timestamps (rather than -1), and we need 2 events (and I would suggest 3, so that we can verify that it doesn't stomp on other ratings.  So something like:

```
RatingBuilder rb = new RatingBuilder().setUserId(420);
rs.add(rb.setItemId(840).setRating(3.5).setTimestamp(10));
rs.add(rb.setItemId(390).setRating(4.5).setTimestamp(20));
rs.add(rb.setItemId(840).clearRating().setTimestamp(30));
```

Also, we need to re-construct the DAO as well, because it takes a copy of the rating list.

After these changes, I think it will be ready to merge! Good work.

elehack:
Read to merge once the tests pass.

elehack:
The AppVeyor failure was an intermittent failure. Rebuilding.

Issue #653, opened 2015-01-09
Add descriptions to CLI --help output


elehack:
Each command line program should have a description in its `--help` output.

Issue #637, opened 2014-11-03
lenskit-eval --version fails


elehack:
Right now, `lenskit-eval --version` does not work:

```
$ ./bin/lenskit-eval --version
Exception in thread "main" java.lang.NullPointerException
	at org.grouplens.lenskit.eval.cli.EvalCLIOptions.lenskitVersion(EvalCLIOptions.java:123)
	at org.grouplens.lenskit.eval.cli.EvalCLIOptions.parse(EvalCLIOptions.java:106)
	at org.grouplens.lenskit.eval.cli.EvalCLI.main(EvalCLI.java:55)
```

`lenskit version` works.

PR #659, opened 2015-01-27
Fixes and improvements for LensKit CLI

elehack:
This clears out the remaining CLI issues for 2.2 (#637, #653).


elehack:
No review needed, will merge when tests pass.

elehack:
And this code doesn't really even get tested, so we'll just merge.

PR #665, opened 2015-02-04
Initial data specification support

elehack:
This implements the specification system and uses it to load data source configurations. LensKit commands now accept data source specifications that describe the text file, etc.

We need to make the text data source more configurable, but this lays the groundwork.


elehack:
Merging. This PR captures the set of changes, if we want post-hoc review.

PR #669, opened 2015-02-12
Add ItemNameDAO

elehack:
This adds an `ItemNameDAO` for generic support for giving items names (e.g. movie titles), along with support in the `recommend` CLI to print recommended item names.

Questionable decisions I made:

- [x] `ItemNameDAO` does not extend `ItemDAO`
- [x] `MapItemNameDAO` implements both `ItemNameDAO` and `ItemDAO`
- [x] `@ItemFile` is reused for the `CSVFileItemNameDAOProvider`, on the premise that using both an item name file and an item ID list file is an unlikely scenario

The first two are affected by the fact that LensKit sort-of implicitly assumes that getItemIds() will return all items that might ever be seen in an event.  This can be false for two reasons:

- in our case, the title file might not have all the movies
- in a database, a new item might be added and rated between the two calls

Guaranteeing that `getItemIds()` will return all item IDs requires heavy database snapshotting. I am very hesitant to require this.

The better fix, IMO, is to make sure LensKit algorithms work fine when `getItemIds()` is missing items. We can test for this with a modification of ML-100K. An algorithm should either ignore unknown items, or augment its storage to accomodate them (either is acceptable behavior).

If we go with this better fix (should create a separate issue for it), then having `MapItemNameDAO` implement `ItemDAO` seems appropriate, and the only question is whether `ItemNameDAO` should extend `ItemDAO`.

### DIFF
--- a/lenskit-cli/src/main/java/org/grouplens/lenskit/cli/InputData.java
+++ b/lenskit-cli/src/main/java/org/grouplens/lenskit/cli/InputData.java
@@ -26,9 +26,12 @@ import net.sourceforge.argparse4j.inf.MutuallyExclusiveGroup;
 import net.sourceforge.argparse4j.inf.Namespace;
 import org.grouplens.lenskit.core.LenskitConfiguration;
 import org.grouplens.lenskit.data.dao.EventDAO;
+import org.grouplens.lenskit.data.dao.MapItemNameDAO;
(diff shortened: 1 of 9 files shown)


elehack:
Some things that need to be done, from offline convo w/ @kluver:

- [ ] Describe how to configure it in the JavaDoc

Issue #504, opened 2014-01-16
Allow SimpleEvaluator to receive algorithms as configuration + name


elehack:
The simple evaluator right now just lets you add `LenskitAlgorithmInstance`s. We need to let the user add a `LenskitConfiguration` and a name.

elehack:
With the 2ed551f backport, this is done.

PR #679, opened 2015-02-25
Add AlgoInstance with name and configuration

ell27:
Add algorithm instance when given name and Lenskit configuration from user.(#504)

### DIFF
--- a/lenskit-eval/src/main/java/org/grouplens/lenskit/eval/traintest/SimpleEvaluator.java
+++ b/lenskit-eval/src/main/java/org/grouplens/lenskit/eval/traintest/SimpleEvaluator.java
@@ -51,6 +51,16 @@ public class SimpleEvaluator implements Callable<Table> {
     }
 
     /**
+     * An algorithm instance constructed with a name and Lenskit configuration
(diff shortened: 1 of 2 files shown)


elehack:
It looks like the problem is a missing import.

lenskit-eval/src/main/java/org/grouplens/lenskit/eval/traintest/SimpleEvaluator.java
```diff
@@ -51,6 +52,16 @@ public SimpleEvaluator() {
     }
 
     /**
+     * An algorithm instance constructed with a name and Lenskit configuration
+     * @param name
+     * @param config Lenskit configuration
+     *
+     */
+    public SimpleEvaluator(String name, LenskitConfiguration config) {
```
elehack: This should be `public SimpleEvaluator addAlgorithm`, instead of just `SimpleEvaluator`; then it should work. Also, coudl we move it down with the other `addAlgorithm` methods?

Thanks! Good work.

elehack:
Perfect! Merging.

PR #682, opened 2015-03-02
Add a 'crossfold' command

elehack:
This adds a `crossfold` command, that crossfolds a data set. It can read the crossfold configuration from a spec, or from command-line arguments. It writes out the data, along with spec files for the train-test data sets.

Unlike the `CrossfoldTask`, this command takes an output directory instead of arbitrary train/test patterns. This is easier to understand, I think.

In order for this to work, I added the `Specifiable` interface for the spec system, that produces a spec from an object.


elehack:
This `Specifiable` thing and paths might be a touch tricky - right now, it outputs full paths. I think that's probably OK, but we might also want relativization capabilities, at least in LensKit 3.

elehack:
I've made `Specifiable` output relative paths. However, the specifications are still all in terms of URIs; I'm not sure we really want that.

Issue #662, opened 2015-02-03
Create time-limited views of binary rating DAOs


elehack:
For the temporal evaluator, we need to be able to create time-limited views of ratings data.

I think that the way to do this is by adding a `createWindowedView(long)` method to `BinaryPackedRatingDAO` that takes a timestamp and returns a new rating DAO that has the same backing storage as the original DAO but only returns ratings before the specified timestamp.

Several things that we need to do here:

- [x] Create the new view method
- [x] Create a suite of tests that create limited views of a fixed set of data and test that all methods correctly limit the ratings they return.
- [x] Implement limiting

For completeness, the view method needs to have defined behavior on a limited view. I think that it would be reasonable to allow views to be limited, but only with a limit before or equal to the current limit (if the current limit is less than the proposed new limit, just reuse the current limit). That way, limited views aren't magically different, and you can't make more ratings available by increasing the limit.

elehack:
Done in #692, improvements moved to #707.

PR #683, opened 2015-03-04
Binary index table improvements

elehack:
This adds tests, comments, and some clarifying refactorings to the binary index table.


elehack:
For linking: this is to help @tajinder-txstate with #662.

Issue #496, opened 2014-01-08
Remove deprecated vector classes


mdekstrand:
Looks like we are going to use vectorz, unless we decide we don't like it for some reason while working on #495.

In that case, we should remove the to-be-deprecated `Vec` class and its friends.

mdekstrand:
This has been done.

PR #686, opened 2015-03-11
Remove deprecated data structures

elehack:
This removes several deprecated data structures and concepts from the LensKit 3 branch:

- Non-sparse vectors (#496)
- Fast iteration of collections and cursors
- `Index`


elehack:
This actually passed, the bad test is due to a spurious push.

Issue #649, opened 2014-12-19
Make event classes non-subclassable concrete classes


elehack:
After conversation with @kluver, we've come to the conclusion that, going forward, event types should be non-subclassable concrete classes with builders and factory functions. Perhaps interfaces, but concrete classes really seem to be the way to go.

elehack:
Completed in #714.

PR #714, opened 2015-04-30
Simplify event classes

elehack:
This implements #649, simplifying event classes to be concrete and non-subclassable.

I am also considering getting rid of `Preference` as an event-level class; that will be a separate PR.


elehack:
If we need the ability to create other objects that work like ratings (such as a scrollable rating or something), I intend to make a `RatingLike` interface that exposes the getters of `Rating`. But we will only do that once we have a use case.

lenskit-core/src/main/java/org/grouplens/lenskit/data/dao/packed/BinaryFormat.java
```diff
@@ -228,22 +227,6 @@ public Rating readRating(ByteBuffer buf) {
         return rb.build();
     }
 
-    /**
-     * Read a rating from a buffer into a mutable rating.
-     * @param buf The buffer to read.
-     * @param rating The rating to populate.
-     */
-    public void readRating(ByteBuffer buf, MutableRating rating) {
```
kluver: Is this function ever used?

lenskit-core/src/main/java/org/grouplens/lenskit/data/dao/packed/BinaryRatingPacker.java
```diff
@@ -252,7 +251,6 @@ private void upgradeRatings(BinaryFormat newFormat) throws IOException {
 
         ByteBuffer oldBuffer = ByteBuffer.allocateDirect(format.getRatingSize());
         ByteBuffer newBuffer = ByteBuffer.allocateDirect(newFormat.getRatingSize());
-        MutableRating scratch = new MutableRating();
```
kluver: Its worth pointing out. I find this code a lot easier to read when its doing "normal java stuff" instead of the mutable optimizations.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
```diff
@@ -20,14 +20,120 @@
...
  */
-public interface Like extends Event {
+public final class Like implements Event, Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private final long userId;
+    private final long itemId;
+    private final long timestamp;
+
+    Like(long uid, long iid, long ts) {
```
kluver: Its been a while since we've bickered about constructors.

Why isn't this public? I would want this to be public.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
```diff
@@ -20,14 +20,120 @@
...
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+import java.io.Serializable;
+
 /**
  * Generic interface representing a unary interaction.  This provides a basic, generic representation
  * of likes, clicks, purchases, and other unary interactions.  Applications may want to implement
  * their own events to represent such interactions.
+ * <p>
+ * This class can be created with {@link LikeBuilder} or the {@link #create(long, long, long)} method.
```
kluver: Not that I think its a big issue, but do we _need_ this many ways to make a like? this seems a little excessive.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
kluver: If memory serves, the argument against using constructors has to do with "what happens when we need to change the object and how its constructed" events should be very simple POJOs. Any significant change to how we build a like is probably precipitated by a change in what is a like, which would be a breaking change anyways. so I guess Im not sure I buy that argument in this case.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
```diff
@@ -20,14 +20,120 @@
...
+    @Override
+    public long getTimestamp() {
+        return timestamp;
+    }
+
+    /**
+     * Create a new builder initialized with a copy of this event's contents.
+     * @return A new builder initialized with this event's data.
+     */
+    public LikeBuilder copyBuilder() {
```
kluver: Isn't there a cloneable interface in java. That makes more sense to me than a copyBuilder, that said, see above comments about not being sure we need a builder anyways.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/LikeBatch.java
```diff
@@ -20,18 +20,102 @@
...
+    }
+
+    @Override
+    public long getItemId() {
+        return itemId;
+    }
+
     /**
-     * Get the number of likes
+     * Get the number of items liked.
```
kluver: I don't think this comment update is accurate. I was under the impression that like batches refer to count type implicit data. The number of items referenced by one likeBatch would therefore be 1, the number of times the user "did the thing" (for any measured action) is what varies.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/LikeBatch.java
```diff
@@ -20,18 +20,102 @@
...
+     * @return The number of items liked.
      */
-    int getCount();
+    public int getCount() {
```
kluver: I know Im going to get pushback on this one, but I want to see what that pushback is.
Do we need getters? if the object is final can't we just have public final attributes like "count?"

kluver: Next function - a good reason to have getters (interface compatibility) not a good reason for final variables to be private.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Rating.java
```diff
@@ -20,23 +20,95 @@
...
+
 /**
- * A rating is an expression of preference for an item by a user.
- *
- * <p>Ratings are equal if they have the same user, item, preference and timestamp.  The hash code
- * of a rating is the hash code of its preference (if it has one), or the user and item ID.
- * Timestamp is <em>ignored</em> in the hash code.  See {@link Ratings#equals(Rating, Rating)} and
- * {@link Ratings#hashRating(Rating)}.</p>
+ * A rating is an expression of preference for an item by a user.  The preference can be {@code null}, in which case
+ * the rating is a removal of a previous rating.
```
kluver: maybe add "commonly referred to as an "unrate""

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
kluver: Also, you don't reference #create(long, long)

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Rating.java
```diff
@@ -44,21 +116,135 @@
...
+
+        @Nullable
+        @Override
+        public Preference getPreference() {
+            return null;
+        }
+
+        @Override
+        public double getValue() throws IllegalStateException {
+            throw new IllegalStateException("Rating has no preference.");
```
kluver: "This rating object has no preference"?

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Rating.java
```diff
@@ -20,23 +20,95 @@
...
+import java.io.Serializable;
+
 /**
- * A rating is an expression of preference for an item by a user.
- *
- * <p>Ratings are equal if they have the same user, item, preference and timestamp.  The hash code
- * of a rating is the hash code of its preference (if it has one), or the user and item ID.
- * Timestamp is <em>ignored</em> in the hash code.  See {@link Ratings#equals(Rating, Rating)} and
- * {@link Ratings#hashRating(Rating)}.</p>
+ * A rating is an expression of preference for an item by a user.  The preference can be {@code null}, in which case
```
kluver: Why do we indirect through a preference object? From what I can see its only to give us a nullable way to code rating vs. unrating?

Also the separation of rating and unrating makes this class hard to understand and more complex that it is clear to me that we need.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Rating.java
```diff
@@ -44,21 +116,135 @@
...
+                                        .toHashCode();
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (obj == this) {
+                return true;
+            } else if (obj instanceof Unrate) {
+                Rating r = (Rating) obj;
+                return new EqualsBuilder().append(user, r.user)
```
kluver: Why use EqualsBuilder here but not other places?

kluver:
Some questions about further ways to simplify this and make it work how someone with a basic java background might expect. My biggest sticking point is the Rating object. There still seems to be some complicatedness in here, and its not at all clear to me why this is as complicated as it is.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/LikeBatch.java
elehack: Yeah - number of times the item was liked would be a better way to term it.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Rating.java
elehack: I have a separate PR in the works that gets rid of the `Preference` indirection (in preparation for entirely getting rid of the `Preference` class). I thought it'd be easier to focus on one set of decisions at a time, so we can get rid of the interface business for `Rating` and then decide on ditching `Preference`.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Rating.java
elehack: I'm trying to standardize on using `EqualsBuilder` basically everywhere, except for trivially-simple cases where the overhead plainly isn't worth it. I had copied `Like` and `LikeBatch` as-is, where they used IntelliJ-generated `equals` methods, I believe; I can change them to use `EqualsBuilder` for consistency.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Rating.java
elehack: I can add those changes to this PR if you'd like to move them through at the same time.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
elehack: There is; it's a bit of a mess, and I don't particularly like it, but it is straightforward for simple classes like this.

But the notion of `copyBuilder()` is also to provide a super-easy way to modify a rating or create another rating/like/like-batch like it. You can do:

```java
Like modified = like.copyBuilder().setItemId(3).build();
```

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
elehack: There are a variety of reasons to make constructors non-public. Many of them do not apply in this case (implementation flexbility, type inference for generic class construction, etc.).

One reason is consistency: if we are going to use the factory method and/or builder patterns for other, more complicated classes in LensKit, it seems to me that it will be easier to work with the code if things are generally done one way. 'LensKit classes are constructed with factory methods' (with the exception of components that participate in dependency injection) is a single rule, rather than having to determine for each class whether it is complicated enough that we decided to use factory methods or builders. And when we write a class, deciding if it will forever be simple enough that we can commit to a public constructor.

I could be dissuaded from this. It may be a good idea to leave the event constructors public, but use factory methods elsewhere.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
elehack: I can add the two-argument reference pretty easily.

Comment above discusses why factory methods instead of constructors. We should perhaps discuss this more.

Even if most code uses the create method, the builder is used as a part of the text file input system. Each builder implements `EventBuilder`, so the definitions of common fields such as `USER` can add the field's value to any kind of event builder.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/LikeBatch.java
elehack: Mostly because public fields are, in general, a code smell, and I don't see a whole lot of cost to just never using them rather than allowing them in certain special cases. There's no run-time overhead to speak of; the JIT will very happily inline `getCount()`, especially since we are making these classes final and concrete (the best-case scenario for method inlining).

lenskit-core/src/main/java/org/grouplens/lenskit/data/dao/packed/BinaryRatingPacker.java
elehack: Good :). I'm wanting to get rid of a lot of that garbage, and re-introduce carefully controlled and better-designed versions of it where performance measurement shows that it is needed.

elehack:
@kluver I've made a bunch of comments now; some stuff I'll address immediately, some we should perhaps have further discussion about. I'm wanting to strike a good balance on the 'best practices' and 'not scary for new devs' scale, and I'm not sure that we've achieved it.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
elehack: One other, trivial, perhaps stupid reason I don't like Clone: the way I usually implement `clone` (directly creating new instance, rather than calling `super.clone()`) results in IntelliJ's warnings framework complaining.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
kluver: I don't feel too strongly on this subject either. 
I think given that we have dependency injection, consistency on factories is out the window.  So Im not sure how much faith I would put on this. Where else do we use factories? (exhaustive list if possible, but if you really have the time it takes to make an exhaustive list I've got some concurrency bugs for you fix)
In the particular case of events I would air on the side of POJO. Its not worth asking people to think about these objects. (At that point, I guess supporting both and using whatever works best in context would be the way to go)

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
kluver: the CloneBuilder as a way to "modify" an immutable object argument is compelling, but not intuitive to me. I definitely say we keep the copy builder then. Is there any way to work this idiom into some sort of example project?

don't implement cloneable. Who would want to clone an imitable object with a well defined equality semantics?

kluver:
Were moving in the right direction. What we want is "all the complexity makes sense" the problem is that what makes sense to some people won't make sense to others, it all depends on how they feel like thinking about LensKit.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
elehack: Agree that cloning an immutable object is a nonsense operation. We can look at putting the `copyBuilder()` into some tutorial docs; I am also thinking about adding `copyBuilder()` to the event interface, so that all events are guaranteed to be copyable in this fashion.

Cloning/copying was more necessary back when we had mutable events, because it was a way to create a long-term savable copy.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
elehack: Is it too confusing to have factory methods, and public constructors, and builders? That would be a defensible design, if perhaps a bit confusing.

Builders have the nice advantage that when you are reading code, you don't have to remember what order things go in, because everything has a name (like Python keyword arguments). For simple events, that isn't a big deal, especially if the code building the object has sensible variable names. But with more than about 4-5 parameters, it can be a bigger deal.

Agree that consistency is already gone with the DI, and that is a decent argument for not trying for a scrupulous consistency here.

I don't have an exhaustive list of factories, and the list is changing anyway because I am starting to use the pattern in more code. It's worth noting that we *don't* have a factory for `LenskitConfiguration`, but do for `LenskitRecommenderEngineBuilder` and `*Loader`. There the difference might be defensible as 'LenskitConfiguration is a "data" object, and the engine builder/loader is a Big Object'. But that seems a bit silly.

Also worth noting: whichever way we go, we are not locked in until we cut a release candidate for LensKit 3.0. We can change all we want before then.

elehack:
@kluver ISTM that the point we're at now is that this is basically ready to merge, we just need to decide how we want to think about constructors, and make sure the `copyBuilder` paradigm gets documented somewhere. Is that consistent with your understanding of the status of this PR?

You probably already noticed, but this PR now includes the dropping of the `Rating`/`Preference` indirection. I think that makes things much simpler. The only place we're now using `Preference` is in the `PackedPreferenceSnapshot` stuff, and I have ideas for retooling that into a more understandable notion of a 'rating matrix'.

lenskit-core/src/main/java/org/grouplens/lenskit/data/event/Like.java
elehack: Two places that use a *lot* of factory methods: the fields and formats for `TextEventDAO` and the table writer package.

Factory methods also make it easy to have utility classes that have factory methods creating commonly-structured instances of other random classes. We can certainly have those and public constructors for things. Perhaps 'factory method all the things' is a foolish consistency.

kluver: Three might be a bit much. I think Ive seen a few reasons why I might like builder objects. I'm not sure we need those and static factory methods. This is a discussion which should go off-line, its too big for this PR. For now, status quo of whatever we have coded.

kluver:
Fully cleaned ratings makes me happy. Also really happy you updated the docs when you switched from null to Nan for unrate.

There are some lingering conversations, but for what this PR set out to do its been done. Lingering convos can be dealt with in the future.

Merging now.

PR #719, opened 2015-05-08
Clean up more warnings and deprecations

elehack:
This goes through some critical Sonar warnings, deletes some deprecated code, and cleans up compiler warnings.

After this set of changes, the LensKit code compiles without any non-documentation compiler warnings on Java 8 with our default compiler flags (`-Xlint -Xlint:-processing`).


elehack:
For review purposes: I don't think there's anything too controversial in here. Things that might feasibly want review:

- enforcement of non-negative damping values
- vararg safety (double-checking my assessment that it's correct)
- I went ahead and dropped some deprecated things; we haven't really talked about how quickly we want to get rid of deprecated code, but I'm in favor of quick cleanup.

elehack:
I am going to merge. @kluver, if you have feedback on the questions, go ahead and raise it post-hoc.

kluver:
Sure.

lenskit-core/src/main/java/org/grouplens/lenskit/data/dao/RatingFile.java
```diff
@@ -1,42 +0,0 @@
...
- *
- * @author <a href="http://www.grouplens.org">GroupLens Research</a>
- * @since 2.1
- */
-@Qualifier
-@Parameter(File.class)
-@Retention(RetentionPolicy.RUNTIME)
-@Target({ElementType.PARAMETER, ElementType.METHOD})
-@Documented
-public @interface RatingFile {
```
kluver: This class does not appear to have been formally deprecated. Based on context that appears more likely to have been an oversight than anything else.

lenskit-core/src/main/java/org/grouplens/lenskit/data/dao/packed/BinaryFormat.java
```diff
@@ -38,7 +38,6 @@
  * @author <a href="http://www.grouplens.org">GroupLens Research</a>
  */
 final class BinaryFormat {
-    public static final byte[] HEADER_MAGIC = "LK".getBytes(Charsets.US_ASCII);
```
kluver: Not used any more?

lenskit-core/src/main/java/org/grouplens/lenskit/data/dao/RatingFile.java
elehack: Yes, probably an oversight. The only component that it configured was deprecated.

lenskit-core/src/main/java/org/grouplens/lenskit/data/history/History.java
```diff
@@ -75,6 +75,8 @@ private History() {}
      * @return An empty history for the user.
      */
     @Nonnull
+    @SafeVarargs
+    @SuppressWarnings("varargs") // method is safe MDE 2015-05-08
     public static <E extends Event> UserHistory<E> forUser(long id, E... events) {
         List<E> list = ImmutableList.copyOf(events);
```
kluver: I'm not going to pretend to fully grock the varargs safety issue, but this looks like the type of code that I understand to be potentially unsafe. It probably depends on whether ImutableList inspects the type of events at runtime, since (as I understand it) events would consider itself a Event array not an E array.

lenskit-core/src/main/java/org/grouplens/lenskit/data/text/DelimitedColumnEventFormat.java
```diff
@@ -51,8 +51,9 @@
     private int headerLines = 0;
 
     @Inject
-    public DelimitedColumnEventFormat(@Nonnull EventTypeDefinition etd) {
+    public DelimitedColumnEventFormat(@Nonnull EventTypeDefinition etd, @ColumnSeparator String delim) {
```
kluver: Could technically be a breaking change if anyone was building this on their own instead of through injection.

lenskit-core/src/main/java/org/grouplens/lenskit/vectors/similarity/PearsonCorrelation.java
```diff
@@ -57,6 +58,7 @@ public PearsonCorrelation() {
 
     @Inject
     public PearsonCorrelation(@SimilarityDamping double s) {
+        Preconditions.checkArgument(s >= 0, "negative damping not allowed");
```
kluver: Is this all the places we use damping, I only counted three, but I feel like there might be more.

lenskit-core/src/main/java/org/grouplens/lenskit/data/history/History.java
elehack: The exact set of warnings I am trying to silence here have to do with a problem called *heap corruption*. Exactly how a heap corruption arises is a complex story involving array covariance and type erasure. However, it is my understanding that it can only happen (as a result of our method) if our method somehow results in the array's contents being changed.

Since the three methods to which I applied this suppression only read the array, never mutate it, I believe that they are safe and the warnings can be safely suppressed. If it would help, I could expand my comments to state that it is safe because the arrays are never modified.

lenskit-core/src/main/java/org/grouplens/lenskit/data/text/DelimitedColumnEventFormat.java
elehack: Yes. I can add a (deprecated?) constructor to fix that.

lenskit-core/src/main/java/org/grouplens/lenskit/data/history/History.java
elehack: Actually, I think there is another case where it arises.

Since the type of `events` is `Object[]` due to type erasure, non-generic code could call `forUser` with some other type of array (either `Object[]`, `StringBuffer[]`). This would be totally fine and unchecked. We will then have a `List<E>` that contains a string buffer, causing some code down the road to attempt to assign a string buffer to an event and break in code that the type checker proved correct.

I *think* that the correct solution in this case is to suppress the varargs warning (because our code is sound), but remove the `SafeVarargs` annotation so that code *calling* this method might still be analyzed for errors. But I might need to read more on `SafeVarargs` to be sure about that conclusion.

elehack: I've gone and asked Stack Overflow about this. http://stackoverflow.com/q/30130262/1385039

lenskit-core/src/main/java/org/grouplens/lenskit/vectors/similarity/PearsonCorrelation.java
elehack: There was one more place I found in a usage search of `@SimilarityDamping`. Fixed in a59a93e.

lenskit-core/src/main/java/org/grouplens/lenskit/data/text/DelimitedColumnEventFormat.java
elehack: Done in 4db4ad3.

lenskit-core/src/main/java/org/grouplens/lenskit/vectors/similarity/PearsonCorrelation.java
kluver: what about meanDamping?

elehack: Good call. Have now fixed those too.

Issue #720, opened 2015-05-08
Define 0 for FP comparison


mdekstrand:
There are various places (e.g. #718) where we need to check if a floating point value is 0. Right now, we do this in several places with `val == 0`. However, using equality comparisons for floating point ops is generally bad.

So it seems to me that it would be good to create a well-defined notion, consistent across LensKit, for ‘is 0’. There are a few things to do to make this happen:
- [x] Implement a stanadard utility function for `isZero()`; I think this can go in a new `MathUtils` class that will collect additional math utilities.
- [x] Use `isZero()` across LensKit.
- [x] Define 0.

My proposed definition of 0: the absolute value is less than `Double.MIN_NORMAL`. This makes all normal floating-point values nonzero, but as soon as the value becomes non-normal (less than 2^-1022), it is considered zero. It seems reasonably non-arbitrary.

We can also add a system property (`lenskit.zero.epsilon`?) to allow the threshold for zero to be redefined.

Issue #718, opened 2015-05-08
Define behavior in MeanVarianceNormalizer when σ=0


mdekstrand:
The `MeanVarianceNormalizer` normalizes items to be z-scores (subtract mean, divide by std dev).

There are two cases in which the standard deviation can be 0:

- the reference vector is empty
- the reference vector is filled with a single value

For both of these cases, the damping must also be 0 to result in a 0 standard deviation. Both cases are also likely to be rare in most data sets.

Note that the normalizer may be used to normalize something other than the reference vector, so in the empty case, it may still be used to normalize values.

What should we do when we encounter a 0 standard deviation? Right now, we just fill values with 0, but that doesn't seem right. There are a few options:

- treat σ=0 as σ=1
- keep as-is (use 0 rather than (x-µ)/σ)
- do nothing (resulting in infinite or NaN values in the normalized/de-normalized vector)

elehack:
Thoughts from @kluver most welcome.

elehack:
I feel like we might be solving a problem that doesn't exist. OTOH, we need to have well-defined behavior, and the current behavior is not well-defined, and edge cases are one thing that hurt reproducibility.

kluver:
I havn't thought the math through carefully, but I think I'm in favor of σ = 1.

If I recall correctly the normalizer will be used in two places: ratings entering the recommender and predictions leaving the recommender.

Entering the recommender is a trivial problem. we should zero center (some algos might assume zero centering and behave in bad, degenerate ways if its not centered), but if there is only 1 or 0 distinct rating values it doesn't really matter.

Exiting the recommender is harder. Obviously we should de-zero center. Since we didn't rescale on the way in, I would say we shouldn't rescale on the way out. In essence we say that z-score normalization degenerates into user-mean normalization when σ is poorly defined.

~Kluver

elehack:
@kluver That is basically my thought as well. We can also emit a warning if it computes a variance of 0 and has no damping in effect - that way the integrator can know that happened. And if they want different behavior, they can apply a damping (in which case we always have a positive numerator, and don't need special care to avoid div-by-0).

So: I think we have agreement that using σ = 1 when the computed variance is 0 *and* there is no damping is a defensible move.

elehack:
@kluver I've put together an implementation of this (and proposed solution for #720) in #721.

PR #721, opened 2015-05-08
Add zero checking

mdekstrand:
This works on #720, adding a zero-checking method and starting to use it in various places.

It also addresses #718, using a standard deviation of 1 if it is 0.


mdekstrand:
I am going to go ahead and merge this, but leave #720 open to track any further discussion about the definition of 0.

Issue #727, opened 2015-05-20
Implement new result API


mdekstrand:
Recommendation results should have a simplified API, that doesn't use our custom sparse vectors and is extensible to provide access to relevant details in a more documented fashion.

[[DetailedResults]] is the starting point for this design, and it is a part of replacing `SparseVector`.

mdekstrand:
We have a new result API!

PR #733, opened 2015-05-21
Create result set APIs

mdekstrand:
This works on #727.

We are not ready to merge this - it is just interfaces. However, I would like to think about interfaces before taking the time to implement.

The idea here is that (almost) all recommendation operations return a *result set*: a (possibly ordered) map of item (or user, for finding related users) IDs to values. Individual algorithms can extend the result set to provide additional details (see [DetailedResults](//github.com/lenskit/lenskit/wiki/DetailedResults)).

For recommendation, the result set will be ordered in decreasing order of score.

This subsumes both `SparseVector` and `List<ScoredId>` for recommendation and prediction results.

### DIFF
--- a/lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
@@ -62,7 +62,7 @@ public interface ItemRecommender {
      * @return The recommended items.
      * @see #recommend(long, int, Set, Set)
      */
-    ResultList<Result> recommend(long user);
+    ResultList recommend(long user);
 
     /**
      * Recommend up to <var>n</var> items for a user using the default exclude
@@ -73,7 +73,7 @@ public interface ItemRecommender {
      * @return The recommended items.
      * @see #recommend(long, int, Set, Set)
      */
-    ResultList<Result> recommend(long user, int n);
+    ResultList recommend(long user, int n);
 
     /**
      * Recommend all possible items for a user from a set of candidates using
@@ -85,7 +85,7 @@ public interface ItemRecommender {
      * @return The recommended items.
      * @see #recommend(long, int, Set, Set)
      */
-    ResultList<Result> recommend(long user, @Nullable Set<Long> candidates);
+    ResultList recommend(long user, @Nullable Set<Long> candidates);
 
     /**
      * Produce a set of recommendations for the user. This is the most general
@@ -107,7 +107,7 @@ public interface ItemRecommender {
      *         decreasing order of score. This is not a hard requirement — e.g.
      *         set recommenders are allowed to be more flexible.
      */
-    ResultList<Result> recommend(long user, int n, @Nullable Set<Long> candidates,
+    ResultList recommend(long user, int n, @Nullable Set<Long> candidates,
                                  @Nullable Set<Long> exclude);
 
     /**
@@ -128,6 +128,6 @@ public interface ItemRecommender {
      *         decreasing order of score. This is not a hard requirement — e.g.
      *         set recommenders are allowed to be more flexible.
      */
-    ResultList<? extends Result> recommendWithDetails(long user, int n, @Nullable Set<Long> candidates,
-                                                      @Nullable Set<Long> exclude);
+    ResultList recommendWithDetails(long user, int n, @Nullable Set<Long> candidates,
+                                    @Nullable Set<Long> exclude);
 }
--- a/lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
@@ -21,17 +21,16 @@
 package org.lenskit.api;
 
 import javax.annotation.Nullable;
+import java.util.List;
 import java.util.Set;
 
 /**
- * Interface for recommending items. Several methods are provided, of varying
- * generality.
+ * Interface for recommending items. This interface provides APIs for both basic (recommend for a given user)
+ * and complex (recommend at most *n* items from a given set to the user) recommendation tasks.
  *
- * The core idea of the recommend API is to recommend <i>n</i> items for a user,
+ * The core idea of the recommend API is to recommend *n* items for a user,
  * where the items recommended are taken from a set of candidate items and
- * further constrained by an exclude set of forbidden items. Items in the
- * candidate set but not in the exclude set are considered viable for
- * recommendation.
+ * further constrained by an exclude set of forbidden items.
  *
  * ## Candidate Items
  *
@@ -62,18 +61,19 @@ public interface ItemRecommender {
      * @return The recommended items.
      * @see #recommend(long, int, Set, Set)
      */
-    ResultList recommend(long user);
+    List<Long> recommend(long user);
 
     /**
-     * Recommend up to <var>n</var> items for a user using the default exclude
+     * Recommend up to `n` items for a user using the default exclude
      * set.
      *
      * @param user The user ID.
-     * @param n    The number of recommendations to return.
+     * @param n    The number of recommendations to return.  Negative values indicate no preference of recommendation
+     *             list size.
      * @return The recommended items.
      * @see #recommend(long, int, Set, Set)
      */
-    ResultList recommend(long user, int n);
+    List<Long> recommend(long user, int n);
 
     /**
      * Recommend all possible items for a user from a set of candidates using
@@ -85,7 +85,7 @@ public interface ItemRecommender {
      * @return The recommended items.
      * @see #recommend(long, int, Set, Set)
      */
-    ResultList recommend(long user, @Nullable Set<Long> candidates);
+    List<Long> recommend(long user, @Nullable Set<Long> candidates);
 
     /**
      * Produce a set of recommendations for the user. This is the most general
@@ -101,14 +101,10 @@ public interface ItemRecommender {
      *                   {@code null}, all items are considered candidates.
      * @param exclude    A set of items to be excluded. If {@code null}, a default
      *                   exclude set is used.
-     * @return A list of recommended items. If the recommender cannot assign
-     *         meaningful scores, the scores will be {@link Double#NaN}. For
-     *         most scoring recommenders, the items will be ordered in
-     *         decreasing order of score. This is not a hard requirement — e.g.
-     *         set recommenders are allowed to be more flexible.
+     * @return A list of recommended items.
      */
-    ResultList recommend(long user, int n, @Nullable Set<Long> candidates,
-                                 @Nullable Set<Long> exclude);
+    List<Long> recommend(long user, int n, @Nullable Set<Long> candidates,
+                         @Nullable Set<Long> exclude);
 
     /**
      * Produce a set of recommendations for the user with additional details. This method functions identically to
--- a/lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
@@ -1,0 +1,133 @@
+/*
+ * LensKit, an open source recommender systems toolkit.
+ * Copyright 2010-2014 LensKit Contributors.  See CONTRIBUTORS.md.
+ * Work on LensKit has been funded by the National Science Foundation under
+ * grants IIS 05-34939, 08-08692, 08-12148, and 10-17697.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.api;
+
+import javax.annotation.Nullable;
+import java.util.Set;
+
+/**
+ * Interface for recommending items. Several methods are provided, of varying
+ * generality.
+ *
+ * The core idea of the recommend API is to recommend <i>n</i> items for a user,
+ * where the items recommended are taken from a set of candidate items and
+ * further constrained by an exclude set of forbidden items. Items in the
+ * candidate set but not in the exclude set are considered viable for
+ * recommendation.
+ *
+ * ## Candidate Items
+ *
+ * By default, the candidate set is the universe of all items the recommender
+ * knows about. The default exclude set is somewhat more subtle. Its exact
+ * definition varies across implementations, but will be the set of items the
+ * system believes the user will not be interested in by virtue of already
+ * having or knowing about them. For example, rating-based recommenders will
+ * exclude the items the user has rated, and purchase-based recommenders will
+ * typically exclude items the user has purchased. Some implementations may
+ * allow this to be configured. Client code always has the option of manually
+ * specifying the exclude set, however, so applications with particular needs in
+ * this respect can manually provide the sets they need respected.
+ *
+ * ## Ordering
+ *
+ * If the recommender has an opinion about the order in which recommendations should be displayed,
+ * the result set will present items in that order.  For many recommenders, this will be descending order
+ * by score; however, this interface does not guarantee a relationship between scores and ordering.
+ *
+ * @compat Public
+ */
+public interface ItemRecommender {
+    /**
+     * Recommend all possible items for a user using the default exclude set.
+     *
+     * @param user The user ID.
+     * @return The recommended items.
+     * @see #recommend(long, int, Set, Set)
+     */
+    ResultList<Result> recommend(long user);
+
+    /**
+     * Recommend up to <var>n</var> items for a user using the default exclude
+     * set.
+     *
+     * @param user The user ID.
+     * @param n    The number of recommendations to return.
+     * @return The recommended items.
+     * @see #recommend(long, int, Set, Set)
+     */
+    ResultList<Result> recommend(long user, int n);
+
+    /**
+     * Recommend all possible items for a user from a set of candidates using
+     * the default exclude set.
+     *
+     * @param user       The user ID.
+     * @param candidates The candidate set (can be null to represent the
+     *                   universe).
+     * @return The recommended items.
+     * @see #recommend(long, int, Set, Set)
+     */
+    ResultList<Result> recommend(long user, @Nullable Set<Long> candidates);
+
+    /**
+     * Produce a set of recommendations for the user. This is the most general
+     * recommendation method, allowing the recommendations to be constrained by
+     * both a candidate set and an exclude set. The exclude set is applied to
+     * the candidate set, so the final effective candidate set is
+     * <var>canditates</var> minus <var>exclude</var>.
+     *
+     * @param user       The user's ID
+     * @param n          The number of ratings to return. If negative, there is
+     *                   no specific recommendation list size requested.
+     * @param candidates A set of candidate items which can be recommended. If
+     *                   {@code null}, all items are considered candidates.
+     * @param exclude    A set of items to be excluded. If {@code null}, a default
+     *                   exclude set is used.
+     * @return A list of recommended items. If the recommender cannot assign
+     *         meaningful scores, the scores will be {@link Double#NaN}. For
+     *         most scoring recommenders, the items will be ordered in
+     *         decreasing order of score. This is not a hard requirement — e.g.
+     *         set recommenders are allowed to be more flexible.
+     */
+    ResultList<Result> recommend(long user, int n, @Nullable Set<Long> candidates,
+                                 @Nullable Set<Long> exclude);
+
+    /**
+     * Produce a set of recommendations for the user with additional details. This method functions identically to
+     * {@link #recommend(long, int, Set, Set)}, except that it may produce more detailed results. Implementations will
+     * return subclasses of {@link ResultList} that provide access to additional details about each recommendation.
+     *
+     * @param user       The user's ID
+     * @param n          The number of ratings to return. If negative, there is
+     *                   no specific recommendation list size requested.
+     * @param candidates A set of candidate items which can be recommended. If
+     *                   {@code null}, all items are considered candidates.
+     * @param exclude    A set of items to be excluded. If {@code null}, a default
+     *                   exclude set is used.
+     * @return A list of recommended items. If the recommender cannot assign
+     *         meaningful scores, the scores will be {@link Double#NaN}. For
+     *         most scoring recommenders, the items will be ordered in
+     *         decreasing order of score. This is not a hard requirement — e.g.
+     *         set recommenders are allowed to be more flexible.
+     */
+    ResultList<? extends Result> recommendWithDetails(long user, int n, @Nullable Set<Long> candidates,
+                                                      @Nullable Set<Long> exclude);
+}
--- a/lenskit-api/src/main/java/org/lenskit/api/ItemScorer.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/ItemScorer.java
@@ -1,0 +1,63 @@
+/*
+ * LensKit, an open source recommender systems toolkit.
+ * Copyright 2010-2014 LensKit Contributors.  See CONTRIBUTORS.md.
+ * Work on LensKit has been funded by the National Science Foundation under
+ * grants IIS 05-34939, 08-08692, 08-12148, and 10-17697.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.api;
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+
+/**
+ * Score items for users.  These scores can be predicted ratings, relevance
+ * scores, purchase probabilities, or any other real-valued score which can be
+ * assigned to an item for a particular user.
+ *
+ * @compat Public
+ */
+public interface ItemScorer {
+    /**
+     * Score a single item.
+     *
+     * @param user The user ID for whom to generate a score.
+     * @param item The item ID to score.
+     * @return The score, or `null` if no score can be generated.
+     */
+    Result score(long user, long item);
+
+    /**
+     * Score a collection of items.
+     *
+     * @param user  The user ID for whom to generate scores.
+     * @param items The item to score.
+     * @return The scores for the items. This result set may not contain all requested items.
+     */
+    @Nonnull
+    ResultMap<Result> score(long user, @Nonnull Collection<Long> items);
+
+    /**
+     * Score a collection of items and potentially return more details on the scores.
+     *
+     * @param user  The user ID for whom to generate scores.
+     * @param items The item to score.
+     * @return The scores for the items. This result set may not contain all requested items.  Implementations that
+     * support additional details will return a subclass of {@link ResultMap} that provides access to those details.
+     */
+    @Nonnull
+    ResultMap<? extends Result> scoreWithDetails(long user, @Nonnull Collection<Long> items);
+}
--- a/lenskit-api/src/main/java/org/lenskit/api/ItemScorer.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/ItemScorer.java
@@ -22,6 +22,7 @@ package org.lenskit.api;
 
 import javax.annotation.Nonnull;
 import java.util.Collection;
+import java.util.Map;
 
 /**
  * Score items for users.  These scores can be predicted ratings, relevance
@@ -48,7 +49,7 @@ public interface ItemScorer {
      * @return The scores for the items. This result set may not contain all requested items.
      */
     @Nonnull
-    ResultMap score(long user, @Nonnull Collection<Long> items);
+    Map<Long,Double> score(long user, @Nonnull Collection<Long> items);
 
     /**
      * Score a collection of items and potentially return more details on the scores.
--- a/lenskit-api/src/main/java/org/lenskit/api/ItemScorer.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/ItemScorer.java
@@ -28,7 +28,7 @@ import java.util.Collection;
  * scores, purchase probabilities, or any other real-valued score which can be
  * assigned to an item for a particular user.
  *
- * @compat Public
+ *  @compat Public
  */
 public interface ItemScorer {
     /**
--- a/lenskit-api/src/main/java/org/lenskit/api/ItemScorer.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/ItemScorer.java
@@ -48,7 +48,7 @@ public interface ItemScorer {
      * @return The scores for the items. This result set may not contain all requested items.
      */
     @Nonnull
-    ResultMap<Result> score(long user, @Nonnull Collection<Long> items);
+    ResultMap score(long user, @Nonnull Collection<Long> items);
 
     /**
      * Score a collection of items and potentially return more details on the scores.
@@ -59,5 +59,5 @@ public interface ItemScorer {
      * support additional details will return a subclass of {@link ResultMap} that provides access to those details.
      */
     @Nonnull
-    ResultMap<? extends Result> scoreWithDetails(long user, @Nonnull Collection<Long> items);
+    ResultMap scoreWithDetails(long user, @Nonnull Collection<Long> items);
 }
--- a/lenskit-api/src/main/java/org/lenskit/api/RatingPredictor.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/RatingPredictor.java
@@ -22,12 +22,14 @@ package org.lenskit.api;
 
 import javax.annotation.Nonnull;
 import java.util.Collection;
+import java.util.Map;
 
 /**
- * Predict user ratings.  A rating predictor is like an {@link ItemScorer}, but its output will be
- * predicted ratings.
+ * Predict user ratings.  A rating predictor is like an {@link ItemScorer}, but its output is scaled or otherwise
+ * transformed for rating prediction.  An item score can be anything that meets the requirement 'higher is better';
+ * a rating prediction can be interpreted as an estimate of the user's expected rating, within the system's stated
+ * range of valid ratings.  Depending on the predictor used, rating predictions may also be quantized.
  *
- * @author <a href="http://www.grouplens.org">GroupLens Research</a>
  * @compat Public
  */
 public interface RatingPredictor {
@@ -50,7 +52,7 @@ public interface RatingPredictor {
      *         not contain all requested items.
      */
     @Nonnull
-    ResultMap predict(long user, @Nonnull Collection<Long> items);
+    Map<Long,Double> predict(long user, @Nonnull Collection<Long> items);
 
     /**
      * Predict the user's preference for a collection of items, potentially with additional details.
--- a/lenskit-api/src/main/java/org/lenskit/api/RatingPredictor.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/RatingPredictor.java
@@ -1,0 +1,65 @@
+/*
+ * LensKit, an open source recommender systems toolkit.
+ * Copyright 2010-2014 LensKit Contributors.  See CONTRIBUTORS.md.
+ * Work on LensKit has been funded by the National Science Foundation under
+ * grants IIS 05-34939, 08-08692, 08-12148, and 10-17697.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.api;
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+
+/**
+ * Predict user ratings.  A rating predictor is like an {@link ItemScorer}, but its output will be
+ * predicted ratings.
+ *
+ * @author <a href="http://www.grouplens.org">GroupLens Research</a>
+ * @compat Public
+ */
+public interface RatingPredictor {
+    /**
+     * Predict a user's rating for a single item.
+     *
+     * @param user The user ID for whom to generate a prediction.
+     * @param item The item ID whose rating is to be predicted.
+     * @return The predicted preference, or {@link Double#NaN} if no preference can be
+     *         predicted.
+     */
+    Result predict(long user, long item);
+
+    /**
+     * Predict the user's preference for a collection of items.
+     *
+     * @param user  The user ID for whom to generate predicts.
+     * @param items The items to predict for.
+     * @return A mapping from item IDs to predicted preference. This mapping may
+     *         not contain all requested items.
+     */
+    @Nonnull
+    ResultMap<Result> predict(long user, @Nonnull Collection<Long> items);
+
+    /**
+     * Predict the user's preference for a collection of items, potentially with additional details.
+     *
+     * @param user  The user ID for whom to generate predicts.
+     * @param items The items to predict for.
+     * @return A mapping from item IDs to predicted preference. This mapping may
+     *         not contain all requested items.
+     */
+    @Nonnull
+    ResultMap<? extends Result> predictWithDetails(long user, @Nonnull Collection<Long> items);
+}
--- a/lenskit-api/src/main/java/org/lenskit/api/RatingPredictor.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/RatingPredictor.java
@@ -50,7 +50,7 @@ public interface RatingPredictor {
      *         not contain all requested items.
      */
     @Nonnull
-    ResultMap<Result> predict(long user, @Nonnull Collection<Long> items);
+    ResultMap predict(long user, @Nonnull Collection<Long> items);
 
     /**
      * Predict the user's preference for a collection of items, potentially with additional details.
@@ -61,5 +61,5 @@ public interface RatingPredictor {
      *         not contain all requested items.
      */
     @Nonnull
-    ResultMap<? extends Result> predictWithDetails(long user, @Nonnull Collection<Long> items);
+    ResultMap predictWithDetails(long user, @Nonnull Collection<Long> items);
 }
--- a/lenskit-api/src/main/java/org/lenskit/api/Recommender.java
+++ b/lenskit-api/src/main/java/org/lenskit/api/Recommender.java
@@ -1,0 +1,65 @@
+/*
+ * LensKit, an open source recommender systems toolkit.
+ * Copyright 2010-2014 LensKit Contributors.  See CONTRIBUTORS.md.
+ * Work on LensKit has been funded by the National Science Foundation under
+ * grants IIS 05-34939, 08-08692, 08-12148, and 10-17697.
+ *
+ * This program is free software; you can redistribute it and/or modify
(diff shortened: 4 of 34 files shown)


mdekstrand:
One additional point of this design: if we adopt something like this, then we can make `lenskit-api` free of both `SparseVector` and fastutil.

mdekstrand:
Some review questions:

- do we like this design direction?
- do we want to extend `Map<Long,Double>`, or provide our own (simpler, read-only) interface with the ability to wrap it in a Map-compatible view? Upon further reflection, I think this would probably be good - result sets do not need to be modifiable.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
```diff
@@ -0,0 +1,133 @@
...
+     *                   {@code null}, all items are considered candidates.
+     * @param exclude    A set of items to be excluded. If {@code null}, a default
+     *                   exclude set is used.
+     * @return A list of recommended items. If the recommender cannot assign
+     *         meaningful scores, the scores will be {@link Double#NaN}. For
+     *         most scoring recommenders, the items will be ordered in
+     *         decreasing order of score. This is not a hard requirement — e.g.
+     *         set recommenders are allowed to be more flexible.
+     */
+    RecResults<? extends ResultEntry> recommendWithDetails(long user, int n, @Nullable Set<Long> candidates,
```
kluver: I kinda like this idea, we could have "with details" recommendation that returns List<Result> for fancy people, and not with details which just returns List<Long> for people who just want a basic recommendation experience.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * ## Ordering
+ *
+ * If the recommender has an opinion about the order in which recommendations should be displayed,
+ * the result set will present items in that order.  For many recommenders, this will be descending order
+ * by score; however, this interface does not guarantee a relationship between scores and ordering.
+ *
+ * @compat Public
+ */
+public interface ItemRecommender {
```
kluver: I assume we will also have an abstract class that does some aggressive delegation to make most people only need to implement one function?

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
mdekstrand: Pretty much. The non-detailed one can also be more efficient by skipping the details, if that's a relevant thing. The default abstract class will delegate everything to with-details, and just wrap or convert the results.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
mdekstrand: Yep, just like we have now.

mdekstrand:
@kluver Thanks. I've pushed up a few changes to implement `ResultList`/`ResultMap`, based on my understanding of current state of mailing list discussion.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Recommend all possible items for a user using the default exclude set.
+     *
+     * @param user The user ID.
+     * @return The recommended items.
+     * @see #recommend(long, int, Set, Set)
+     */
+    ResultList<Result> recommend(long user);
```
kluver: ResultList<Result> instanceof List<Result>?

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
```diff
@@ -0,0 +1,133 @@
...
+     *                   {@code null}, all items are considered candidates.
+     * @param exclude    A set of items to be excluded. If {@code null}, a default
+     *                   exclude set is used.
+     * @return A list of recommended items. If the recommender cannot assign
+     *         meaningful scores, the scores will be {@link Double#NaN}. For
+     *         most scoring recommenders, the items will be ordered in
+     *         decreasing order of score. This is not a hard requirement — e.g.
+     *         set recommenders are allowed to be more flexible.
+     */
+    ResultList<Result> recommend(long user, int n, @Nullable Set<Long> candidates,
```
kluver: Theoretically these "without details" versions could simply return List<Long> for people who just want plain old java.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
```diff
@@ -0,0 +1,133 @@
...
+     *                   {@code null}, all items are considered candidates.
+     * @param exclude    A set of items to be excluded. If {@code null}, a default
+     *                   exclude set is used.
+     * @return A list of recommended items. If the recommender cannot assign
+     *         meaningful scores, the scores will be {@link Double#NaN}. For
+     *         most scoring recommenders, the items will be ordered in
+     *         decreasing order of score. This is not a hard requirement — e.g.
+     *         set recommenders are allowed to be more flexible.
+     */
+    ResultList<? extends Result> recommendWithDetails(long user, int n, @Nullable Set<Long> candidates,
```
kluver: Is the '<? extends result>' here buying us anything? Its been a while since I looked over the question mark corners of generics.

kluver:
Its worth noting that Im not reading the documentation on this pass. That will probably want its own pass once we get the other details happy.

lenskit-api/src/main/java/org/lenskit/api/Result.java
```diff
@@ -0,0 +1,43 @@
...
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.api;
+
+public interface Result {
```
kluver: I'm not clear why this isn't an override-able class.

lenskit-api/src/main/java/org/lenskit/api/Result.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Get the ID for the result.
+     * @return The user or item ID for this result.
+     */
+    int getId();
```
kluver: long?

lenskit-api/src/main/java/org/lenskit/api/ResultList.java
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @param type The result type to cast to.  It is always valid for this type to be a supertype of `E`; the method
+     *             will also succeed if every non-null result is an instance of this type (or one of its subtypes).
+     * @param <T> The result type.
+     * @return A result list statically typed to contain results of type `T`.
+     * @throws IllegalArgumentException if not all results can be cast type type `T`.
+     * @throws UnsupportedOperationException if the result list is mutable (very rare, only used for intermediate
+     * working objects).
+     */
+    <T extends Result> ResultList<T> castResults(Class<T> type);
```
kluver: `T extends E`?

lenskit-api/src/main/java/org/lenskit/api/Result.java
kluver: One point I suppose is sub-interfaces which would allow multiple sub-interfaces to be implemented by a class.

lenskit-api/src/main/java/org/lenskit/api/ResultList.java
kluver: We need to think carefully and probably prototype patterns for accessing detailed result types. Its not clear that this API is useful without knowing how we want this used.

lenskit-api/src/main/java/org/lenskit/api/ResultMap.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * A set of results from a recommender operation.  Recommendation results conceptually map item
+ * (or user, for a few operations) IDs to corresponding scores, and have some order over their IDs.
+ * Depending on the operation that produced the result set, that order may be arbitrary, but
+ * operations that find things (recommendation, related item and user queries) will usually
+ * return results in order from most to least relevant or wanted.
+ *
+ * @param <E> The entry type (parameterized to make extension cleaner).
+ */
+public interface ResultMap<E extends Result> extends Map<Long,E>, Iterable<E> {
```
kluver: getTopN might not be a bad method to add to this. Its kinda mixing stuff, but it will probably be a common pattern.

kluver:
My biggest concerns are 

  1. do we want to offer functions that go to base types like List<Long> or Map<Long, Double> for people who want a more normal java experience?
  2. We need to think carefully and probably have sample code around how to use type casting to get extra details.

lenskit-api/src/main/java/org/lenskit/api/Result.java
mdekstrand: Yep.

lenskit-api/src/main/java/org/lenskit/api/ResultList.java
mdekstrand: `T extends Result` is what we want, so that we can up-cast.

The canonical use of this will be to implement the non-detailed `recommend` method:

```java
public ResultList<T> recommend(long user, ...) {
    return recommendWithDetails(user, ...).castResults(Result.class);
}
```

However, if we mandate that result lists *always* be immutable, then it is safe to just do an unchecked cast.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
mdekstrand: Yes - it allows detailed result classes returned by an implementation to implement e.g. `ResultList<ItemItemResult>`. Since Java does not allow us to declare that `ResultList` is covariant in its type parameter, this is our only way to make that kind of subtyping type-check.

mdekstrand: In general, I don't like wildcards in return types, and prefer to restrict their use to parameter types. However, Java does not give us any options here.

mdekstrand: I shouldn't say *no* options. There is another option - require individual results to be cast, rather than allowing the scorer/recommender to be cast to gain type-safe automatic access to the results.

lenskit-api/src/main/java/org/lenskit/api/Result.java
mdekstrand: Primary reason is probably because I still instinctively reach for interfaces everywhere.

In this case, I think it buys us benefit by allowing implementations/extensions complete freedom in how they implement the result interface. Want to be a flyweight, consisting only of an index into some bizarre structure containing the real results? Cool.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
kluver: Thats an option that doesn't offend me. That said `<? extends Result>` works well enough for me as well.

lenskit-api/src/main/java/org/lenskit/api/Result.java
kluver: I think this decision co-mingles with how we want to support casting these stupid objects to get access to their soft gooey detail rich explanations.

lenskit-api/src/main/java/org/lenskit/api/ResultMap.java
mdekstrand: How about a `getTopN` static utility method somewhere that can operate over any kind of map? It can be built on the `TopNAccumulator` logic, possibly even live there. That way implementations don't have to carry the implementation around.

lenskit-api/src/main/java/org/lenskit/api/ResultList.java
kluver: Im still not clear on two things.

1. isn't up casting perfectly legal? I can always go `Object foo = new Bar()` without any problem (I think). Does the generics somehow break this property?

2. I don't really care how "internals" deal with casting and types. Not as deeply as I care about client code. Lets get concrete. Im writing a website (say booklens) and I want to push some confidence measure out to my users. I have some function that turns neighborhood sizes into text labels like "trustworthy" or whatever. 

Ideally starting from a RatingPredictor instance (as opposed to a subclass unless we want to make that more common...), how do I iterate over the predictions in a clean(-ish) way and get the neighborhood size, pass it to my function, and then do something with the results?

lenskit-api/src/main/java/org/lenskit/api/ResultMap.java
kluver: that last point more valid if these were objects and we could provide one reference impl and not worry about it.

kluver: Just noticed that, why arn't _these_ classes?

mdekstrand:
@kluver wrote:
> do we want to offer functions that go to base types like List or Map for people who want a more normal java experience?

Is the fact that `ResultList` extends `List` enough to achieve this?

> We need to think carefully and probably have sample code around how to use type casting to get extra details.

Yes.

There are a couple of ways that we could go. What I was originally envisioning is that you would usually do something like this:

```java
ItemItemScorer scorer = rec.get(ItemItemScorer.class);
ItemItemResultMap results = score.scoreWithDetails(user, items);
for (ItemItemResult result: results) {
    System.out.format("%d: %.3f (%d neighbors)\n", result.getId(), result.getScore(), result.getNeighborCount());
}
```

We could also go with an individual result casting strategy:

```java
ItemItemScorer scorer = rec.get(ItemItemScorer.class);
ResultMap results = scorer.scoreWithDetails(user, items);
for (Result result: results) {
    ItemItemResult iir = (ItemItemResult) result;
    System.out.format("%d: %.3f (%d neighbors)\n", iir.getId(), iir.getScore(), iir.getNeighborCount());
}
```

Now, things get a little funny when we're dealing with more complicated components.

```java
FallbackItemScorer scorer = rec.get(FallbackItemScorer.class);
ResultMap results = scorer.scoreWithDetails(user, items);
for (Result result: results) {
    // we can get score
    result.getScore();
    FallbackScorerResult fsr = (FallbackScorerResult) result;
    // primary:
    Result pr = fsr.getPrimaryResult();
    // fallback:
    Result fb = fsr.getFallbackResult();
}
```

Now, we have a problem: we really want the details from the primary result, which is something like an `ItemItemResult`. In an ideal world, `FallbackItemScorer` would be parameterized in its primary and fallback scorers, and their result types, and would return a parameterized `FallbackScorerResult`. However, generic types and the dependency injector don't get along, so we have avoided making injectible components such as item scorers generic. But we will, for any sufficiently interesting result type, need to deal with casting individual results; if we don't have to for the top-level results, we will have to for nested results.

Therefore, it seems to me that since we will often need to cast individual results (because most interesting configurations will use some kind of a wrapper, like a simple rating predictor or a top-n item recommender or a fallback item scorer), being able to achieve type safety by casting entire scorers or result maps/sets to simplify the 30% case of just looking at outermost results is not worth it.

We could introduce an 'intelligent' casting API, `as`, that converts the result type and returns `null` when conversion fails:

```java
ItemScorer scorer = rec.getItemScorer();
ResultMap results = scorer.scoreWithDetails(user, items);
for (Result result: results) {
    ItemItemResult iir = result.as(ItemItemResult.class);
    assert iir != null; // if this fails, then our item scorer is not item-item and we are bad
    System.out.format("%d: %.3f (%d neighbors)\n", iir.getId(), iir.getScore(), iir.getNeighborCount());
}
```

For most basic result types, such as item-item results, the `as` method would just check the class and, if it's a superclass of the result implementation type, return the result, otherwise return `null`.

This can get interesting for more complicated results. The `SimplePredictionResult`, which will contain the prediction and provide the underlying scorer's result as a detail, can return the underlying result if its type matches the `as` type:

```java
public <R extends Result> R as(Class<R> type) {
    if (type.isInstance(this)) {
        return type.cast(this);
    } else {
        return rawScore.as(type);
    }
}
```

The `TopNItemRecommender` can just use its underlying scorer's results as its result type, and they can be type-cast with `as`.

The fallback scorer can try itself, then its left, and then its right implementations. So if you have a `FallbackScorer` that has a primary of an item-item and a baseline of a personalized mean, you can still get the item-item details with the exact same code above - the `as(ItemItemResult.class)` call will basically be 'if this thing is an item-item result, find it.'

Now, if we go this route, there is room for things to get surprising if you have particularly complicated configurations (e.g. a hybrid of a few item-item recommenders). But I think there are a few mitigating factors:

- If you are doing sophisticated things with results, at some point you do need to know how your recommender is structured, and be more careful. There is no avoiding that.
- Each composite result type should have clearly-defined search behavior, that is designed to minimize surprise.
- We could further help by having two casting methods: `as`, which just does flat casting, and `find` that does the recursive search for an appropriate result type.

kluver:
part 1: markdown ate some text. lemme escape it right: `List<Long>` and `Map<Long>` yes the inheritance is good, Im just wondering if everyone will want to deal with the Result objects.

(because your feedback on comment part 2 will take a while to read I won't get to it for a while.)

mdekstrand:
@kluver I agree with adding methods to view the results as such a list. I think I also like the idea of making them the return type for the non-detailed methods - with good design for the result type implementations, it could even improve efficiency for the common case.

kluver:
I don't know that I like the `FallbackScorerResult` but I don't think thats the point. (Im more interested in a mix of types and forcing people to write more robust result processing code)

I like the direct use with no casting when people want to hardcode implementations. We will have to make it clear and obvious how to do that (get is a strange and magic function to be sure). 

We will definitely need to think through the type casting variants. the "as" function seems compelling. I might be temped to say lets design that out. I personally would say lets make it work with interfaces as well so that we could create "equivalence classes" (or somesuch nonsense) of meaningful outputs (all knn algorithms, for example, can provide neighborhood sizes). 

Does `as` actually give us much over the java `isntanceof` keywords? because if it doesn't we should just example code using normal java type casting concepts.

I like this. It looks like we can do it with pretty good leverage of standard java type stuff. See comment about the confusing type-thingy in the rec list.

Lets do this.

mdekstrand:
@kluver wrote:
> Does as actually give us much over the java isntanceof keywords? because if it doesn't we should just example code using normal java type casting concepts.

In the flat version, it gives us a little DRY. Instead of `instanceof` followed by a `cast`, both mentioning the same class, it lets us have a single mention followed by a null check.

In the recursive search version, it gives us, the full recursive search.

> We will definitely need to think through the type casting variants. the "as" function seems compelling. I might be temped to say lets design that out. I personally would say lets make it work with interfaces as well so that we could create "equivalence classes" (or somesuch nonsense) of meaningful outputs (all knn algorithms, for example, can provide neighborhood sizes).

Working with interfaces should be pretty easy.

> Lets do this.

Will do. I think I can implement some simple compatibility wrappers so that we don't have to migrate all of the implementations in a single pull request - we'll see.

mdekstrand:
@kluver wrote:
> I like the direct use with no casting when people want to hardcode implementations. We will have to make it clear and obvious how to do that (get is a strange and magic function to be sure). 

I'm working on designing out the exact classes, and so far this requirement is resulting in some rather cumbersome use of generics.

mdekstrand:
@mdekstrand wrote:
> @kluver wrote:
> > I like the direct use with no casting when people want to hardcode implementations. We will have to make it clear and obvious how to do that (get is a strange and magic function to be sure).
>
> I'm working on designing out the exact classes, and so far this requirement is resulting in some rather cumbersome use of generics.

I think I have a design for this now. Each custom implementation type can provide its own method that returns a List<CustomResultType>. Since the caller will be getting a specific implementation type anyway, we don't need the method to be specified in a general interface.

The `as`/`find` methods can convert result types in other cases.

mdekstrand:
OK, this now has a complete pass at result set APIs and the new public APIs.

lenskit-api/src/main/java/org/grouplens/lenskit/RecommenderEngine.java
```diff
@@ -39,5 +39,5 @@
      *
      * @return A recommender ready for use.
      */
-    public Recommender createRecommender();
```
kluver: Uhh... I'm not sure I follow this change.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
```diff
@@ -0,0 +1,133 @@
...
+     *                   {@code null}, all items are considered candidates.
+     * @param exclude    A set of items to be excluded. If {@code null}, a default
+     *                   exclude set is used.
+     * @return A list of recommended items. If the recommender cannot assign
+     *         meaningful scores, the scores will be {@link Double#NaN}. For
+     *         most scoring recommenders, the items will be ordered in
+     *         decreasing order of score. This is not a hard requirement — e.g.
+     *         set recommenders are allowed to be more flexible.
+     */
+    ResultList recommendWithDetails(long user, int n, @Nullable Set<Long> candidates,
```
kluver: I don't know that this function is necessary. Since the Java side API is identical to recommend (the one above it) it certainly isn't buying us Java side generality. And if the user wants the extra detail version I would assume they would be class casting anyways so this wouldn't need to be in the interface.

Also I can't imagine wanting to implement an algorithm that has separate implementations for both the with and without details version instead of pointing one function at the other. Do you think that is a likely to happen?

Last time I saw this I had thought maybe the other functions could simply return `List<Long>`, making "withDetails" separate on a java side but I think the to simple list functionality makes more sense (or at lease just as much sense) on the result list.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
```diff
@@ -0,0 +1,133 @@
...
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.api;
+
+import javax.annotation.Nullable;
+import java.util.Set;
+
+/**
+ * Interface for recommending items. Several methods are provided, of varying
+ * generality.
```
kluver: "Several methods are provided, of varying generality"

While not technically wrong, Im not sure this really communicates anything special. Maybe something like: "This class provides apis for both basic (recommend for a given user), and complex (recommend at most n items from a given set to the user) recommendation tasks "

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * The core idea of the recommend API is to recommend <i>n</i> items for a user,
+ * where the items recommended are taken from a set of candidate items and
+ * further constrained by an exclude set of forbidden items. Items in the
```
kluver: What is the purpose of the excluded item set? Why separate it from candidates instead of asking people to perform a set difference on their own?

kluver: Maybe give an example:
"an exclude set of forbidden items, which can be used to prevent recommending items the user has already rated"

kluver: Arrg, never mind, Explained in next paragraph. Maybe don't have the last sentence if it will be explained in the next paragraph?

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
```diff
@@ -0,0 +1,133 @@
...
+     * @see #recommend(long, int, Set, Set)
+     */
+    ResultList recommend(long user);
+
+    /**
+     * Recommend up to <var>n</var> items for a user using the default exclude
+     * set.
+     *
+     * @param user The user ID.
+     * @param n    The number of recommendations to return.
```
kluver: document negative n behavior here

lenskit-api/src/main/java/org/lenskit/api/ItemScorer.java
```diff
@@ -0,0 +1,63 @@
...
+     * @return The score, or `null` if no score can be generated.
+     */
+    Result score(long user, long item);
+
+    /**
+     * Score a collection of items.
+     *
+     * @param user  The user ID for whom to generate scores.
+     * @param items The item to score.
+     * @return The scores for the items. This result set may not contain all requested items.
```
kluver: Maybe document null response (as above) for some items.

lenskit-api/src/main/java/org/lenskit/api/ItemScorer.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Score a collection of items and potentially return more details on the scores.
+     *
+     * @param user  The user ID for whom to generate scores.
+     * @param items The item to score.
+     * @return The scores for the items. This result set may not contain all requested items.  Implementations that
+     * support additional details will return a subclass of {@link ResultMap} that provides access to those details.
+     */
+    @Nonnull
+    ResultMap scoreWithDetails(long user, @Nonnull Collection<Long> items);
```
kluver: As above (recommender), so long as this shares API with its partner function above, Im not sure if we need it.

lenskit-api/src/main/java/org/lenskit/api/RatingPredictor.java
```diff
@@ -0,0 +1,65 @@
...
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.api;
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+
+/**
+ * Predict user ratings.  A rating predictor is like an {@link ItemScorer}, but its output will be
```
kluver: maybe "buts its output is guaranteed to be scaled for rating prediction" Or some other promise style words to make this more clearly different.

lenskit-api/src/main/java/org/lenskit/api/RatingPredictor.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Predict the user's preference for a collection of items, potentially with additional details.
+     *
+     * @param user  The user ID for whom to generate predicts.
+     * @param items The items to predict for.
+     * @return A mapping from item IDs to predicted preference. This mapping may
+     *         not contain all requested items.
+     */
+    @Nonnull
+    ResultMap predictWithDetails(long user, @Nonnull Collection<Long> items);
```
kluver: as above.

lenskit-api/src/main/java/org/lenskit/api/RecommenderEngine.java
```diff
@@ -0,0 +1,43 @@
...
+ * <p>
+ * Recommender engines can be shared across threads or sessions; they are
+ * long-lived objects that should be built once.  Recommenders are opened "per-request"
+ * or per thread and provide access to the actual recommendation components.
+ * </p>
+ *
+ * @compat Public
+ * @see Recommender
+ */
+public interface RecommenderEngine {
```
kluver: Didn't I already review this class?

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
```diff
@@ -0,0 +1,72 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Main entry point for accessing recommender components.  A recommender object
+ * is effectively a recommender <i>session</i>: it is a per-thread or per-request
+ * object, likely connected to a database connection or persistence session, that
+ * needs to be closed when the client code is finished with it.
+ *
+ * <p>The various methods in this class return {@code null} if the corresponding
```
kluver: javadoc html confuses me frequently. Do you ever need to close the <p> tag here?

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Get the recommender's item scorer.
+     * @return The item scorer for the configured recommender, or {@code null} if item scoring is not supported.
+     */
+    @Nullable
+    ItemScorer getItemScorer();
+
+    /**
+     * Close the recommender.  This closes underlying resources such as database collections. Components retrieved
+     * from the recommender cannot be used once the recommender is closed.
```
kluver: Any particular exception we should expect if this happens?

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    @Nullable
+    ItemRecommender getItemRecommender();
+
+    /**
+     * Get the recommender's item scorer.
+     * @return The item scorer for the configured recommender, or {@code null} if item scoring is not supported.
+     */
+    @Nullable
+    ItemScorer getItemScorer();
```
kluver: Is this really a distinct task from RatingPredictor?

(On the other hand, does it really cost us anything to have this here?)

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
kluver: Additional thought for consideration. Does this only need to exist / have to exist to support composing recommenders, Like the `RatingPredictor` that simply wraps an `ItemScoerer`?

lenskit-api/src/main/java/org/lenskit/api/Result.java
```diff
@@ -0,0 +1,77 @@
...
+     */
+    boolean hasScore();
+
+    /**
+     * Convert this result to the specified type, if possible.
+     *
+     * @param type The desired result type.
+     * @param <T> The desired result type.
+     * @return The result as a result of type `T`, or `null` if the type cannot be cast.  This may be done by casting
+     * this type, or by unwrapping wrapper result types, but it will not search multiple alternatives.
```
kluver: The concept of "searching multiple alternatives" isn't clear to me.

lenskit-api/src/main/java/org/lenskit/api/Result.java
```diff
@@ -0,0 +1,77 @@
...
+    @Nullable
+    <T extends Result> T as(@Nonnull Class<T> type);
+
+    /**
+     * Attempt to view this result as another type, searching through alternatives if necessary.
+     *
+     * @param type The type to find.
+     * @param <T> The desired result type.
+     * @return The first result of type `T` found by searching this result and all results it contains.  Each type of
+     * result will define some search order for the purpose of defining the 'first' result of type `T`.
```
kluver: I think you are thinking of the case of composite/hybrid recommenders, right? I wonder if that isn't some form of overkill for this situation. Cast it to the composite result type and then pull it out yourself.

lenskit-api/src/main/java/org/lenskit/api/ResultMap.java
```diff
@@ -0,0 +1,49 @@
...
+     * View this result set as a map from longs to doubles.
+     * @return A map view of this result set.
+     */
+    Map<Long,Double> scoreMap();
+
+    /**
+     * View the results as a set.
+     * @return The result set.
+     */
+    Set<Result> resultSet();
```
kluver: Does this buy us anything? For a second there I was thinking iteration, but thats already in the interface.

lenskit-core/src/main/java/org/lenskit/LenskitRecommender.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    @Override
+    public ItemRecommender getItemRecommender() {
+        return get(ItemRecommender.class);
+    }
+
+    @Override
+    public void close() {
+        // TODO Implement closing
```
kluver: Should this todo concern me?

lenskit-core/src/main/java/org/lenskit/results/AbstractResult.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * The default implementation simply calls {@link #as(Class)}.
+     */
+    @Override
+    public <T extends Result> T find(@Nonnull Class<T> type) {
+        return as(type);
```
kluver: This is what I expect most Lenskit users mental model of as/find to be (should they even have a mental model).

lenskit-core/src/main/java/org/lenskit/results/AbstractResult.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Create a hash code builder, populated with the ID and score.  Subclasses can use this as a starting point for
+     * building a hash code.
+     *
+     * @return A hash code builder that has the ID and score already appended.
+     */
+    protected HashCodeBuilder startHashCode() {
```
kluver: Cool piece of API here. I havnt' seen this before, but I kinda like it.

I notice that you don't actually implement HashCode and Equals.

lenskit-core/src/main/java/org/lenskit/results/BasicResult.java
```diff
@@ -0,0 +1,94 @@
...
+     * Create a new basic result.
+     * @param id The result ID.
+     * @param score The result score.
+     * @see Results#create(long, double)
+     */
+    public BasicResult(long id, double score) {
+        super(id, score);
+    }
+
+    private void writeObject(ObjectOutputStream out) throws IOException {
```
kluver: Do these have to do with Serializable?

lenskit-core/src/main/java/org/lenskit/results/LenskitResultList.java
```diff
@@ -0,0 +1,31 @@
...
+package org.lenskit.results;
+
+import it.unimi.dsi.fastutil.longs.LongList;
+import org.lenskit.api.ResultList;
+
+/**
+ * LensKit refinement of the general result list type.
+ */
+public interface LenskitResultList extends ResultList {
+    LongList idList();
```
kluver: I was wondering where the more optimal classes were going to come in.

lenskit-core/src/test/groovy/org/lenskit/LenskitRecommenderEngineTest.groovy
```diff
@@ -0,0 +1,538 @@
+/*
```
kluver: Im going to trust you and only briefly skim the tests.

kluver:
I have a small number of documentation writing-level comments. 

The as/find functions (being separate) confuses me a little, and I expect will confuse others. I'm worried we are overenginerring for very complex result types, which would be almost unsable anyways.

The "withDetails" as a member of the interface seems unnecessary to me (unless there is something noteworthy that is different in the Java)

lenskit-api/src/main/java/org/grouplens/lenskit/RecommenderEngine.java
mdekstrand: Cleanup and churn - interface methods are implicitly public, so `public` keyword is unnecessary.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
mdekstrand: @kluver wrote:
> Also I can't imagine wanting to implement an algorithm that has separate implementations for both the with and without details version instead of pointing one function at the other. Do you think that is a likely to happen?

For simple-to-compute scalar values, it is not useful.

However, it could be useful to report *very* detailed results, such as have the item-item CF implementation report the actual neighbors used for each recommendation. These could then be extracted in a metric and analyzed. Doesn't increase the time, but does increase the size of the result objects and the lifetime of intermediate objects (which may e.g. push them out of edenspace into a longer-lived GC realm...). I don't expect writing two separate implementations; rather, it's more likely to create a common implementation function that is called by each, with a flag determining how much bookkeeping to do.

Also, each detail adds memory overhead in the result list.

kluver: A fair argument.

Im still not sure if this needs to be part of the interface. It feels to me like any use of this implementation will necessarily have knowledge of the implementation.

lenskit-api/src/main/java/org/lenskit/api/RecommenderEngine.java
mdekstrand: lol we have 2 of them now.

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
mdekstrand: Since we're using Markdown now, I'll just delete the tag.

lenskit-api/src/main/java/org/lenskit/api/RecommenderEngine.java
kluver: `.grouplens.*` vs. not. It took me a while to get that...

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
mdekstrand: Any component capable of detecting a 'stuff's been closed' condition should throw `IllegalStateException`. Components are not required to check for this, and may have no way of knowing. But LensKit does not guarantee that objects are usable after the recommender has been closed.

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
mdekstrand: Good question, and I'm not entirely happy with the distinction. The idea is that we can rank, etc. on pre-clamped scores. I have thought about removing `ItemScorer` from the 'public' API, and making it a very common Lenskit-internal feature.

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
kluver: Might be worth documenting some of that in some way.

lenskit-api/src/main/java/org/lenskit/api/ResultMap.java
mdekstrand: Good question, and I don't know.

lenskit-core/src/main/java/org/lenskit/LenskitRecommender.java
mdekstrand: No. LK3 will require recommenders to be closed, but we do not yet have the infrastructure to actually close them. So the API is in there as a stub, so new code can be written correctly even before closing does anything.

lenskit-core/src/main/java/org/lenskit/results/BasicResult.java
mdekstrand: Yes. AbstractResult is not Serializable, but BasicResult is, so it needs some help.

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
mdekstrand: The reason I put it in the interface is so that wrapper and hybrid scorers can just pass their inner recommenders' details through unchanged, and propagate whether or not the caller cares about details.

Might be overkill, but seems useful.

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
kluver: One semi persuasive point in my oppinion is that I don't think we have any item Scorers that arn't rating predictors (right now). one way to fix that would be to implement some things that we clearly believe arn't ratingPredictors so we can point at them and be happy.

lenskit-api/src/main/java/org/lenskit/api/ResultMap.java
kluver: get rid of it?

lenskit-core/src/main/java/org/lenskit/results/BasicResult.java
kluver: Mind throwing a comment
`// Serialization support functions`
Or something before the block of code?

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
kluver: IMHO, it feels like overkill as is, but necissary for hybrids. My personal preferred implementation would be if other functions returned direct `Lists<Long>` or `Maps<Long,Double>`. Then we have near-0 overhead most of the time, and variable overhead in those few cases that we care about extra details.

lenskit-core/src/main/java/org/lenskit/results/BasicResult.java
mdekstrand: Sure. Can even make it an IntelliJ region so it folds.

lenskit-core/src/main/java/org/lenskit/results/AbstractResult.java
mdekstrand: Correct - any subclass will need to do that, and add any additional fields it carries around. I just wanted to make it easy to have consistent behavior over the common fields.

lenskit-api/src/main/java/org/lenskit/api/Result.java
mdekstrand: Yeah. The idea is 'I know there's an ItemItemResult somewhere in there - go find it for me', so that the detail-extracting code doesn't need to know the entire structure of the recommender (and is resilient to modest changes in configuration, such as adding a denormalizing wrapper).

The idea is that if the result just wraps something (e.g. a SimpleRatingPredictor wrapping an ItemItemScorer will return a RatingPrediction that wraps an ItemItemResult), then `as` will get the wrapped object. But if there are *two* wrapped objects (e.g. a linear blend hybrid), then `as` will **not** try to pick one whereas `find` will.

`as` may be enough, as you can `as` to get the hybrid result and then go poking around in it.

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
mdekstrand: Configure ItemItemScorer with the SimilaritySumNeighborhoodScorer, and voila!

lenskit-api/src/main/java/org/lenskit/api/Result.java
kluver: I would recommend that. The hybrid specific class can even either implement as in a funny way, or have find as a class specific.

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
kluver: bam, got me. that weakens my resolve a little bit. 

how about ItemScorer with RatingPredictor as a subinterface. Maybe one that reports its own range or something (just so its mroe than just a tag interface)?

lenskit-api/src/main/java/org/lenskit/api/ItemRecommender.java
mdekstrand: My only hesitancy about this is that it makes the 'medium' case (item recommendation with scores) invoke the full overhead of the details.

mdekstrand: But I think I will do it anyway.

kluver: You do have a fair point. I think one (possibly not great) way to get around this would be a "verbose" flag at constructor time.

This is all still pre-release. We have a chance to try it out and decide if it isn't working.

gtrefs:
I am not that deep into the internals. But would it not be easier to actually be more functional? I read some parts of your discussion and hat a look at the wiki page. But in the end, the outcome of the recommendation is a data structure, right? Why not provide functions like ```map(Function<E, T> f)```, ```flatMap(Function<E, ResultList<T> f)``` and ```filter(Predicate<E> p)```.
* ```map(Function<E, T> f)```: Convert Result
* ```flatMap(Function<E, ResultList<T> f)```: Apply next recommender.
* ```filter(Predicate<E> p)```: Filter result

```java
new ItemItemScorer().predict(1, new ArrayList(){{add(1);add(2);add(3);}}).filter(score -> score < 0.4).flatMap(...);
```

mdekstrand:
@gtrefs Thanks for the input.

There are a lot of things I like about a functional model like you are describing. And actually, I would like configuring LensKit to look more like that, possibly in LensKit 3.1.

But currently, it would be a very significant change from LensKit's current design based on dependency-injected components. I have some long-term ideas for how to build a functional-style API that generates DI configuration under the hood; I'm not sure if that's a good idea or not.

The DI model gives us significant ability to algorthmically reason over the structure of a recommender algorithm, to do things like automatically identify opportunities for sharing components between recommender instances in an evaluation and provide good, automatic support for some of the nuances of integrating with web applications in an algorithm-independent fashion. It also puts all of the algorithm structure into a configuration file, and application code knows nothing but how to ask for recommendations.

It might be that the tradeoffs of going to a more direct, functional algorithm implementation and configuration style are worth it. I know that DI has been one of the things that has been confusing and complicated for some potential users. Applications can easily enough write a recommender service class that calls the LensKit algorithms.

So far, I have been trying to take the same core ideas that LensKit already has and make them clearer and easier to work with. I'm open to a more thorough rethink, but I think that's a bigger conversation, perhaps for the mailing list. Going in the direction of functionally composing pieces would move LensKit from being a 'framework' to being a library for building (and evaluating?) recommenders. Not a bad plan (and there's a lot to be said for libraries vs. frameworks). Just... a lot of work, and with some nontrivial costs in recovering some of automatic efficiency support. It'd also basically force us to Java 8 (we currently still support 7, based on a poll of LensKit users that I ran in December).

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
mdekstrand: LensKit 1 had that design. I don't remember all the reasons we switched away, but one of them was to decouple rating range clamping from item scoring (so we didn't duplicate the clamping in each scorer), and make it so that scorers could be rating predictors or not based on how they were configured, and we don't have to have two different classes to provide the correct type-level indications.

mdekstrand:
OK, I think I've addressed all of @kluver's comments that required code changes.

Planning to add LenskitItemScorer and friends that return Fastutil interfaces, but I think that can be done later.

lenskit-api/src/main/java/org/lenskit/api/Recommender.java
kluver: complicated.

kluver:
On a quick skim the changes we discussed seem to be present. pending tests good to merge.

Doesn't Java have implementations of functional primitives related to the core collections (part of guava maybe?) At least for predict and recommend outputs a user can always use those for expressive power by using the "with details" results. The issue is that the "with details" functions may take longer to run and require more ram than their primitive versions. so supporting the basic output types (recommend directly to item ids for example) allows more efficient use of our API.

Issue #726, opened 2015-05-20
Provide fine-grained parallelism of builds in evaluator


mdekstrand:
Currently, the evaluator's paralleism is constrained by a very coarse view of how reusable components can be built in parallel.

If instead we replace sharable component values with fork-join tasks, that can be `join`ed by multiple threads, then fork-join will handle our parallelism dependencies itself, and additional builds can proceed as soon as shared components are built.

Subtask of #573.

PR #734, opened 2015-05-21
Synchronize the NodeInstantiator

mdekstrand:
It seems that, in rare cases, we may be triggering the WeakHashMap live-lock discussed here: https://java.net/jira/browse/JAVASERVERFACES-2544

This synchronizes our use of InjectionContainer to avoid that live-lock.

### DIFF
--- a/lenskit-core/src/main/java/org/grouplens/lenskit/inject/NodeInstantiator.java
+++ b/lenskit-core/src/main/java/org/grouplens/lenskit/inject/NodeInstantiator.java
@@ -73,7 +73,11 @@ public abstract class NodeInstantiator implements Function<DAGNode<Component,Dep
 
         @Override
         public Object instantiate(DAGNode<Component, Dependency> node) throws InjectionException {
-            return container.makeInstantiator(node).instantiate();
+            Instantiator inst;
+            synchronized (container) {
+                inst = container.makeInstantiator(node);
+            }
+            return inst.instantiate();
         }
     }
 }


kluver:
I will test and merge tonight or tomorrow.

mdekstrand:
@kluver Great!

Went ahead and made the change b/c, even if it isn't the problem or we don't have time to debug, our current usage is clearly incorrect.

kluver:
This seems to have made concurrency problems much worse. I will look into it more as time is available. please do not merge.

mdekstrand:
@kluver Ugh.

It may not be worth too much investigation if we're going to do #726 and change a bunch of things in this space.

kluver:
You might not be wrong, but I am quite stubborn and want to fix this and figure out why its happening. Since the bug is in the integration with Lenskit and Grapht I can't really do a great of debugging, but here are two observations:

 * When it does lock, in this version all but two threads are _blocked_ (instead of their past waiting cycle) the threads are blocked waiting at the new synchronization point. the two running threads are in the WeakHashMap again, One appears to be on line 78 making the instantiator, and the other is on line 80 using that instantiator.

 * I tested a version where instantiation is part of the same synchronzation block. As might be expected this was a bad idea. I didn't see locking but I also noticed that at most 2 threads seemed to be running at any point of time. 

Thoughts?

mdekstrand:
Not a lot. I have also seen this bug crop up just running LensKit's integration tests on my laptop, although very infrequently. I have not tried a heavily parallel run with this code yet to see its impact.

My Next Attempt at a fix would be to modify Grapht to synchronize access to the weak map in `InjectionContainer`, and see if that fixes the problem without killing our parallelism/deadlocking.

mdekstrand:
@kluver I am working on some new design that should fix this by other means.

mdekstrand:
Will not merge, since we are getting rid of the weak hash map in the cache.

Issue #737, opened 2015-05-28
Create crossfolding Gradle task


mdekstrand:
After #736 and in support of #296, we need to make a Gradle task for crossfolding.

This will serve as an initial point for implementing the Gradle support for specification-based tasks.

PR #743, opened 2015-06-11
Implement 'crossfold' Gradle task

mdekstrand:
This implements the 'crossfold' Gradle task (#737). Includes tests.

### DIFF
--- a/lenskit-core/src/main/java/org/lenskit/util/DynamicVariable.java
+++ b/lenskit-core/src/main/java/org/lenskit/util/DynamicVariable.java
@@ -1,0 +1,85 @@
+/*
+ * LensKit, an open source recommender systems toolkit.
+ * Copyright 2010-2014 LensKit Contributors.  See CONTRIBUTORS.md.
+ * Work on LensKit has been funded by the National Science Foundation under
+ * grants IIS 05-34939, 08-08692, 08-12148, and 10-17697.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.util;
+
+import java.util.Deque;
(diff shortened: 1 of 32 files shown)


mdekstrand:
This code is odd and Gradle-y enough that I don't expect others to have much opinion of it, so I'll go ahead and merge. @kluver, if you do want to review, feel free to do so post-hock.

mdekstrand:
This PR does give us some useful common infrastructure for building the rest of the Gradle plugin. I'll be building on and refining that as I write additional plugins.

lenskit-core/src/main/java/org/lenskit/util/DynamicVariable.java
```diff
@@ -0,0 +1,85 @@
...
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.util;
+
+import java.util.Deque;
+import java.util.LinkedList;
+
+/**
+ * A dynamic variable, manging scoped changes of value.  These values are **not** thread-local.
```
kluver: Im going to need an explanation of why you reinvented scoping.

lenskit-gradle/src/main/groovy/org/lenskit/gradle/LenskitExtension.groovy
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Each property has a default value; the LensKit plugin also examines the project properties for properties starting
+ * with `lenskit.` to initialize the extension properties.  This allows you to override property defaults on the
+ * command line:
+ *
+ * ```
+ * ./gradlew evaluate -Plenskit.threadCount=10
+ * ```
+ */
+public class LenskitExtension {
```
kluver: Here and/or in LenskitTask seems like a good place to put logging controls.

mdekstrand: I agree completely. LenskitExtension and/or LenskitTask will be the central places for configuring things that are common to all LensKit commands.

lenskit-core/src/main/java/org/lenskit/util/DynamicVariable.java
mdekstrand: Occasionally, it is useful to have vintage-Lisp-style dynamically-scoped variables, so that you have a global-ish variable but change its value for the duration of running some block of code.  The specification context is how specification handlers get things like the current base URI; when I introduced a way to set the base URI in the specification language, I needed a way to change the base URI for all objects configured during the course of configuring the immediate object whose base URI was changed.  SpecificationContext has a mutually recursive relationship with the individual specification handlers; it calls the handler, and provides a reference to itself, which the handler then uses to build sub-objects.

The dynamic variable lets me do exactly that.

However, now that I have tried to explain it, it occurs to me that I could just create a fresh context with a different base URI and pass that, and skip this dynamic variable backdoor. So I will make this balderdash go away.

some days you just gotta implement ancient Lisp in Java. And then realize you don't.

Issue #739, opened 2015-05-28
Create rating packing Gradle task


mdekstrand:
We need to create a rating packing Gradle task to make it easy to pack ratings.

mdekstrand:
We have one now, in #746.

PR #746, opened 2015-06-16
Implement PackRatings Gradle task

mdekstrand:
This creates a PackRatings Gradle task (#739).

### DIFF
--- a/lenskit-cli/src/main/java/org/lenskit/cli/util/Logging.java
+++ b/lenskit-cli/src/main/java/org/lenskit/cli/util/Logging.java
@@ -58,10 +58,11 @@ public final class Logging {
         logging.addArgument("--log-level")
                .type(String.class)
                .metavar("LEVEL")
-               .help("include logging messages at LEVEL in log file");
-        logging.addArgument("-d", "--debug")
-               .action(Arguments.storeTrue())
-               .help("include debug logging in console output");
+               .help("include logging messages at LEVEL in log output");
+        logging.addArgument("--log-file-level")
+               .type(String.class)
+               .metavar("LEVEL")
+               .help("include logging messages at LEVEL in log file (defaults to --log-level value)");
         logging.addArgument("--debug-grapht")
                .action(Arguments.storeTrue())
                .help("include debug output from Grapht");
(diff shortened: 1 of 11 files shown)


mdekstrand:
@kluver This has new logging config support too.

lenskit-cli/src/main/java/org/lenskit/cli/util/Logging.java
```diff
@@ -72,10 +73,15 @@ public static void configureLogging(Namespace options) {
...
             return;
         }
-        boolean debug = options.getBoolean("debug");
         boolean debugGrapht = options.getBoolean("debug_grapht");
         File logFile = options.get("log_file");
 
+        String lstr = options.getString("log_level");
+        Level logLevel = Level.toLevel(lstr, Level.INFO);
+        String lfstr = options.getString("log_file_level");
+        Level logFileLevel;
```
kluver: strange looking code style here

lenskit-cli/src/main/java/org/lenskit/cli/util/Logging.java
```diff
@@ -103,20 +119,18 @@ public static void configureLogging(Namespace options) {
...
-                filter.setLevel("INFO");
-                filter.start();
-                console.addFilter(filter);
-            }
-        } else if (debug) {
-            root.setLevel(Level.DEBUG);
-        } else {
-            root.setLevel(Level.INFO);
+
+            root.setLevel(logFileLevel);
```
kluver: II might be wrong here, but is seems like the rood.setLevel on 110 and 133 can both be reached. That seems strange, but not necessarily wrong, to me (for lack of knowing the logging configuration framework well enough to know better).

lenskit-cli/src/main/java/org/lenskit/cli/util/Logging.java
```diff
@@ -91,10 +97,20 @@ public static void configureLogging(Namespace options) {
         console.setEncoder(consolePat);
         root.addAppender(console);
 
+        // figure out the root appender's level
+        Level rootLevel = logLevel;
+        if (logFileLevel.toInt() < logLevel.toInt()) {
+            rootLevel = logFileLevel;
```
kluver: It feels to me like it might be worth saving this if for when we know if there is a logFile. There is no reason to step down for a lower log file level and then filter if the log file is disabled right?

lenskit-cli/src/main/java/org/lenskit/cli/util/Logging.java
mdekstrand: Not sure what you're getting at - the extra 'str' variables?

kluver: no having logFileLevel declared and then set on the next line, instead of keeping it on one line like you do for logLevel.

mdekstrand: Ahh. Yeah, that's weird, not sure why I did that.

lenskit-cli/src/main/java/org/lenskit/cli/util/Logging.java
mdekstrand: I think that this line is wrong. I will remove it.

lenskit-cli/src/main/java/org/lenskit/cli/util/Logging.java
mdekstrand: Doing this too.

PR #748, opened 2015-06-19
Specification beans

mdekstrand:
This rips out the old specification system in favor of one based on 'specification beans' that live in a `lenskit-specs` module.

The reason is to make the definition of the specification formats encoded in Java classes, but still in an extensible fashion (although the extensibility is not entirely finished and will be fleshed out more as needed).  This means we don't have a lot of 'what did I name that field in the JSON parser?' thinking; we can just code against the spec beans, and they will be serialized with Jackson.


mdekstrand:
I am going to merge this once tests have passed; there's some funny business with the way specs work, but I think it's better to move forward and keep refining over the LK3 cycle than try to have detailed reviews of slightly-weird code right now.

PR #753, opened 2015-08-07
funksvd refactor vector library 

neblo:
Skimmed over the code once again. I am pretty sure it is all in order. The only thing to note is that the there was not a function that returns the sum of all the elements in a realVector, so I wrote up a simple function to do so.

### DIFF
--- a/lenskit-svd/src/main/java/org/grouplens/lenskit/mf/funksvd/FunkSVDItemScorer.java
+++ b/lenskit-svd/src/main/java/org/grouplens/lenskit/mf/funksvd/FunkSVDItemScorer.java
@@ -22,8 +22,7 @@ package org.grouplens.lenskit.mf.funksvd;
 
 import it.unimi.dsi.fastutil.longs.LongSet;
 import it.unimi.dsi.fastutil.longs.LongSortedSet;
-import mikera.vectorz.AVector;
-import mikera.vectorz.Vector;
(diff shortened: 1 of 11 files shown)


lenskit-svd/src/main/java/org/grouplens/lenskit/mf/funksvd/FunkSVDItemScorer.java
```diff
@@ -165,34 +164,34 @@ public void score(long user, @Nonnull MutableSparseVector scores) {
         scores.set(estimates);
 
         if (!ratings.isEmpty() && rule != null) {
-            AVector updated = Vector.create(uprefs);
+            //RealVector updated = uprefs.copy();
             for (int f = 0; f < featureCount; f++) {
-                trainUserFeature(user, updated, ratings, estimates, f);
+                trainUserFeature(user, uprefs, ratings, estimates, f); // changed updated to uprefs
```
mdekstrand: We still want to use `updated` - we don't want to update the vector stored in the model.

mdekstrand:
OK, it looks good other than the `updated`/`uprefs` issue. Fix that (check out your refactor branch, make a new commit, push that branch, and the PR will automatically update), and we'll be good to merge.

mdekstrand:
Although now that we've done this, I've also discovered LA4J, another library that looks cleaner. But let's go ahead and merge this, and we'll think on LA4J.

Issue #747, opened 2015-06-16
Support automatically chaining data sets in Gradle


mdekstrand:
We need to be able to tell Gradle tasks to use the output of other Gradle tasks as data sets.

PR #754, opened 2015-08-07
Route data sources through Gradle tasks

mdekstrand:
This augments the Gradle plugin to address #747, allowing data sources to automatically be detected and routed through tasks so that a LensKit Gradle task can take another task as input. It adds support for this to `Crossfold` (consumer) and `PackRatings` (supplier) to test it. 

It also adds a test to make sure that LensKit tasks automatically depend on Java compilation if they are running from the project's classpath.


mdekstrand:
I think this is basically sound; we may want to change some things as I move forward with additional tasks, but experience will show us that. @kluver (or anyone else), feel free to post-hoc.

PR #756, opened 2015-08-14
Add Python support to Appveyor builds

mdekstrand:
This modularizes the AppVeyor builds and adds Python support to them, based on appveyor/ci#359.


ctaggart:
I noticed that the `-q` doesn't appear to be working:
image

I'm also seeing this with our builds and will log a bug with conda.

mdekstrand:
@ctaggart Yeah, I noticed that as well. Thanks for filing a bug about it!

PR #757, opened 2015-08-14
New packed data structure code

mdekstrand:
This does a few things:

- creates a `LongKeyIndex`, like the old key domain but read-only and without masks
- implements packed sets and maps on top of `LongKeyIndex`
- provides initial implementation of vector arithmetic on top of long-to-double maps


mdekstrand:
This code will allow us to retain a number of the performance benefits of the old sparse vector and associated packed data structures, with significantly simpler code.

Issue #769, opened 2015-08-28
Move and rename cursors


mdekstrand:
We need to move cursors into `lenskit-core`, since the data structures package is going away. While we are at it, I would like to consider renaming them, because ‘cursor’ is a somewhat overloaded term (that I may actually want to use for something else, that's really more like a database cursor).

I have found a class in the Java standard library that is very similar to our cursors in most respects: [DirectoryStream](http://docs.oracle.com/javase/8/docs/api/java/nio/file/DirectoryStream.html) (OK, all respects, but C3PO)

I think that it would be reasonable to rename `Cursor` to `ObjectStream`, and move it to `o.l.util.streams` (or `o.l.io.streams`?). This would also make it consistent with our DAO naming: a `streamEvents` call will return an `ObjectStream`. There's a slight overloading with Java 8's streams, but Java already has that between DirectoryStream and Stream. I don't think it is a significant problem.

mdekstrand:
Done in #772.

mdekstrand:
ping @kluver - if you have any objections to this naming, or alternate suggestions, it'd be good to hear them early :)

kluver:
I don't have any alternative suggestions. Ive never been entirely clear why
these arn't just iterators, but I suppose the closing requirement is a
pedantic but meaningful separation because it means that these can't be
dropped into normal iteration idoms.

I also don't understand why iterators don't implement iterable. But that's
neither here nor there.
ObjectStream sounds as good as anything else.

~kluver

mdekstrand:
@kluver Ok.

I've also retooled `ObjectStream` a so it behaves more like an input stream (has a `readObject` method), but also implements `Iterable` so that it can be iterated conveniently.

Issue #715, opened 2015-05-04
Refactor PackedPreferenceSnapshot into RatingMatrix


mdekstrand:
Our `PackedPreferenceSnapshot` is really a randomly-ordered ratings matrix. I think we should double down on this, and treat it as a rating matrix rather than our own strange concept.

So I propose that we rename `PreferenceSnapshot` and associated classes to `RatingMatrix`; we can have one or more implementations (I think leaving `RatingMatrix` as an interface to allow for multiple implementations is appropriate - one implementation can just re-use a binary packed rating DAO, if it has no unrate or rerate events).

There are a few things to do:

- [ ] Rename interface `PreferenceSnapshot` to `RatingMatrix`
- [ ] Rename implementation(s)
- [ ] Rename package (to `org.lenskit.data.matrix`?)
- [ ] Replace `IndexedPreference` with a `RatingMatrixEntry` class or interface defined in the same package as `RatingMatrix`
- [ ] Make sure we have `getUserRatings` and `getItemRatings` methods that are well-defined and reasonably efficient

For the last piece - efficient per-user and per-item ratings - we can implement maps of user/item IDs (or indexes) to lists of the indexes of that user's ratings (similar to the binary packed rating DAO). The rating matrix should still be randomly ordered, so that the efficient operation is iterating over ratings in a randomly-permuted fashion for iterative learning. Long-term, we might want the ability to re-tune the implementation for different use cases.

mdekstrand:
Done in #772.

PR #772, opened 2015-09-05
Move a lot of data around

mdekstrand:
This moves around a lot of data access code and cleans it up.

- Moves Cursors to be ObjectStreams (#769)
- Creates the Ratings Matrix (#715)
- Unify key indexes, removing separate ID Index Mapping concept
- Delete old `Preference` concept, instead having ratings and rating matrix entries that implement `Preference`

This is a lot of churn, but few logical changes. Yay for tests. Naming things is the biggest potential question on this big changeset.


mdekstrand:
Biggest outstanding question is whether `ObjectStream` is really a good name (#769). It's overloaded but precedented, my feelings are complicated.

Issue #732, opened 2015-05-21
Review and revise core API classes


mdekstrand:
We need to review the core API classes (`ItemRecommender`, `RatingPredictor`, etc.). This is for two reasons:

- removing sparse vectors (#730) in favor of a new result API (#727) will make changes in what the APIs look like.
- the classes and APIs are not necessarily how we really want them

I think the following are, conceptually, probably about right:

- `ItemRecommender`
- `ItemScorer`
- `RatingPredictor`

I think we need to reconsider:

- `GlobalItemRecommender` could be come `RelatedItemFinder`
- `GlobalItemScorer` needs a better name, related to scoring items with respect to each other.

Then, for orthogonality, we can add:

- `RelatedUserFinder`
- A user-to-user scoring function

mdekstrand:
This is partially done by #733. Still need to decide on global/related APIs.

kluver:
On it. probable turn around: Thursday.

mdekstrand:
With #790, this is now done.

PR #790, opened 2015-10-20
Rename global scoring interfaces

mdekstrand:
This finishes off #732, removing the remaining old APIs (global scoring and recommendation) and replacing them with new-style APIs.

The new classes are `ItemBasedItemScorer` and `ItemBasedItemRecommender`. They have methods `recommendRelatedItems` and `scoreRelatedItems`.

It seems feasible to use them to identify most-similar items or users in many situations.

@kluver, would like a double-check on naming and docs.


lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
```diff
@@ -29,62 +26,68 @@
...
 
 /**
- * The interface for recommendation based on the items only. The difference from {@link
- * ItemRecommender} is that the input is only the item or list of items instead of user specific
- * information. This interface can be used to provide a Find Similar Items / People Also Liked
- * feature.
+ * Recommends items that go with a set of reference items. This interface is distinguished from
+ * {@link ItemRecommender} in that it uses a set of reference items instead
+ * of a user as the basis for computing scores. This component can be used to provide a Find Similar Items
+ * or People Also Liked feature.
```
kluver: Not 100% sure, (since this note applies to all algorithms) that it depends on which algorithm/configuration whether it is reasonable to use this for similar items. Example:

consider recommender on amazon which is fed in a mid-range microwave. Similar items are other microwaves, good basket recommendations are probably toasters and other kitchenware. Both would use this API, but the meanings are different.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
```diff
@@ -29,62 +26,68 @@
...
  *
  * @see ItemRecommender
  * @author <a href="http://www.grouplens.org">GroupLens Research</a>
  * @compat Public
  * @since 0.10
  */
-public interface GlobalItemRecommender {
+public interface ItemBasedItemRecommender {
```
kluver: I like the "item based" prefix.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
kluver: One follow-up point to this - we have a precedent of separating classes by task, scorers are separated from rating predictors even though the interface is essentially the same.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
```diff
@@ -29,62 +26,68 @@
...
     /**
      * Recommend all possible items for a basket of items using the default exclude set.
```
kluver: This appears to be the first use of the word basket. I think if we want this word we should put it in the top level doc, otherwise I kind of like just calling it a set of reference items.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
```diff
@@ -29,62 +26,68 @@
...
      *
-     * @param items The items value.
+     * @param basket The reference basket.
      * @return The recommended items.
-     * @see #globalRecommend(Set, int, Set, Set)
+     * @see #recommendRelatedItems(Set, int, Set, Set)
      */
-    List<ScoredId> globalRecommend(Set<Long> items);
```
kluver: Do we have something to take a userId look up Likes or all rated items or something and then feed into one of these classes?

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
```diff
@@ -29,62 +26,68 @@
...
+    List<Long> recommendRelatedItems(Set<Long> basket);
```
kluver: should we use a longSet or are we happy with the easier-to-read, but box and unboxy Set<Long>?

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
```diff
@@ -29,62 +26,68 @@
...
 
     /**
      * Recommend up to <var>n</var> items for a basket of items using the default exclude set.
```
kluver: We need to make a decision between default exclude set or no exclude set. I seem to remember the online discussion Scott and I favored no exclude set (or at least a method with no exclude set) over undefined default exclude set.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
```diff
@@ -29,62 +26,68 @@
...
      *
-     * @param items      The items value.
-     * @param candidates The candidate set (can be null to represent the universe).
-     * @return The recommended items.
-     * @see #globalRecommend(Set, int, Set, Set)
+     * @param basket     The reference basket.
+     * @param n          The number of ratings to return. If negative, no specific size is requested.
+     * @param candidates A set of candidate items which can be recommended. If {@code null}, all
+     *                   items are considered candidates.
+     * @param exclude    A set of items to be excluded. If {@code null}, a default exclude set is
```
kluver: Im not sure if I want null to be default or empty for exclude. If we keep the concept of a default exclude, then I think null should be that (even if absent parameter means empty) since empty sets are pretty easy to make.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
```diff
@@ -29,62 +26,68 @@
...
+     *                   used.
+     * @return A list of recommended items.
      */
-    List<ScoredId> globalRecommend(Set<Long> items, @Nullable Set<Long> candidates);
+    List<Long> recommendRelatedItems(Set<Long> basket, int n, @Nullable Set<Long> candidates,
+                                     @Nullable Set<Long> exclude);
 
     /**
-     * Produce a set of recommendations for the item. This is the most general recommendation
+     * Produce a set of recommendations for the item, with details. This is the most general recommendation
```
kluver: Im not sure you get to have two functions say "this is the most general recommendation method" prehaps "this is the most powerful"?

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemScorer.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param items  The list of items to score.
+     * @return A mapping from item IDs to predicted preference. This mapping may
+     * not contain all requested items.
+     */
+    @Nonnull
+    Map<Long, Double> scoreRelatedItems(@Nonnull Collection<Long> basket,
+                                        @Nonnull Collection<Long> items);
+
+    /**
+     * Score items in a vector based on a collection of items (e.g. a shopping basket).
```
kluver: there is no vector anymore.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemScorer.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @param basket The items to use as the query.
+     * @param items  The items to score.
+     */
+    ResultMap scoreRelatedItemsWithDetails(@Nonnull Collection<Long> basket, Collection<Long> items);
```
kluver: Talk about resultMap / with detail-ness in this documentation

lenskit-cli/src/main/java/org/lenskit/cli/commands/GlobalRecommend.java
```diff
@@ -25,16 +25,17 @@
 import net.sourceforge.argparse4j.impl.Arguments;
```
kluver: Should this class still be named GlobalRecommend?

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
```diff
@@ -29,62 +26,68 @@
...
      */
-    List<ScoredId> globalRecommend(Set<Long> items);
+    List<Long> recommendRelatedItems(Set<Long> basket);
 
     /**
      * Recommend up to <var>n</var> items for a basket of items using the default exclude set.
      *
-     * @param items The items value.
+     * @param basket The reference basket.
      * @param n     The number of recommendations to return.
```
kluver: document <0 means no limit.

lenskit-core/src/main/java/org/lenskit/basic/TopNItemBasedItemRecommender.java
```diff
@@ -89,21 +84,15 @@ protected LongSet getDefaultExcludes(LongSet items) {
...
         }
 
-        ScoredItemAccumulator accum = new TopNScoredItemAccumulator(n);
-        for (VectorEntry pred : scores) {
-            final double v = pred.getValue();
-            accum.put(pred.getKey(), v);
-        }
-
-        return accum.finish();
+        return Results.newResultList(Results.scoreOrder().greatestOf(scores, n));
```
kluver: I havn't seen this. Part of me likes the readability, another part of me questions ordering and truncation being part of the same parameter.

lenskit-core/src/main/java/org/lenskit/util/collections/LongUtils.java
```diff
@@ -134,6 +134,20 @@ public static Long2DoubleFunction asLong2DoubleFunction(final Map<Long,Double> m
     }
 
     /**
+     * Create a map that maps a group of items to the same value.
+     * @param keys The keys.
+     * @param value The value.
+     * @return A map that contains all `keys`, mapping each of them to `value`.
+     */
+    public static Long2DoubleMap constantDoubleMap(Set<Long> keys, double value) {
+        // TODO Implement this using a flyweight wrapper
```
kluver: is this todo recorded somewhere not-in-the-code

lenskit-knn/src/main/java/org/lenskit/knn/item/ItemItemItemBasedItemScorer.java
```diff
@@ -21,28 +21,34 @@
...
 import java.util.Collection;
+import java.util.List;
 
 /**
  * Score items based on the basket of items using an item-item CF model.
  *
  * @author <a href="http://www.grouplens.org">GroupLens Research</a>
  */
-public class ItemItemGlobalScorer extends AbstractGlobalItemScorer {
+public class ItemItemItemBasedItemScorer extends AbstractItemBasedItemScorer {
```
kluver: I don't know how to get around this name, but I do want to point out that it contains the word "Item" three times in a row and four times overall.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
mdekstrand: You can get this from the `UserEventDAO`.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
mdekstrand: Currently, `lenskit-api` for LK3 does not use Fastutil. Open to change that, but it does lock our clients into using fastutil.

Internally, we use factory methods to cast or wrap sets to unboxed sets and use Fastutil interfaces.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
mdekstrand: Agree. If `null` is empty, then we have no good to represent default.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
mdekstrand: Hmm. Yes, we should decide (and apply the same decision to `ItemRecommender`).

I like the idea that saying 'recommend(user)` just Does The Right Thing in the simple case. Code can replace `ItemRecommender` to redefine what 'right thing' is (might even introduce a new component to compute default candidate and exclude sets, so that it's easier for applications to redefine Do The Right Thing - a `CandidateIdentifier`).

lenskit-cli/src/main/java/org/lenskit/cli/commands/GlobalRecommend.java
mdekstrand: Good question. I don't know?

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
kluver: Yah, but I could imagine an item scorer that wraps a ItemBasedItemScorer using the DAO for lookup.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
kluver: cool.

lenskit-core/src/main/java/org/lenskit/basic/TopNItemBasedItemRecommender.java
mdekstrand: It's also a little less efficient than our accumulators, I think. But it's really short and concise and code I didn't have to debug. Guava is fun!

> another part of me questions ordering and truncation being part of the same parameter.

Could you clarify your concern here a bit?

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
kluver: I think we need something, since "right thing" and "default set" are nebulous and undefined, which makes them unsuitable for relying on IMHO. Personally I would rather have it work in a way that matches my intuition as a programmer. 

One other point to consider is current clients might _rely_ on this behavior from our current API end points.

the "add a component" answer might be the best, its a little clunky feeling to me, but it does make the concept of default sets well defined and easy to reason about.

lenskit-knn/src/main/java/org/lenskit/knn/item/ItemItemItemBasedItemScorer.java
mdekstrand: Yes, it does.

It sounds like something Doctor Donna might have said as her brain was snapping from time-lord-ish-ness.

We have an outstanding inconsistency, too: our item-item CF `ItemScorer` is called `ItemItemScorer` rather than `ItemItemItemScorer`.

Way to work around this problem: introduce a notion of item similarity (other than the one used for computing the matrices?) and pull this out to be a `SimilarityItemBasedItemScorer` (since that's really what it is), and have item-item CF provide an item-item similarity that pulls from the model.

lenskit-core/src/main/java/org/lenskit/basic/TopNItemBasedItemRecommender.java
kluver: In my mind ordering concerns and limit concerns should be two parameters. That said, if this is guava stuff ignore my concerns.

lenskit-knn/src/main/java/org/lenskit/knn/item/ItemItemItemBasedItemScorer.java
kluver: explicitly talking about item similarity might work. Otherwise the name isn't wrong, its just goofy. Also houw about `ItemBasedKNNItemBasedItemScorer`. its got all the Item and even more filler.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
mdekstrand: If we implement a candidate identifier component, we can then document things as follows:

- The candidate/exclude-set-free methods call the complete method with candidate and exclude sets of `null`.
- The default recommender implementations pass their candidate and exclude sets, as-is, to a `findCandidates(long, LongSet, LongSet)` on `UserBasedCandidateIdentifier` or `findCandidates(LongSet, LongSet, LongSet)` on `ItemBasedCandidateIdentifier` method.
- The default `CandidateIdentifier` implements the following logic:

    - If the candidate set is `null`, use the full set of items from the `ItemDAO`
    - If the exclude set is `null`, use the reference set (for by-item) or the set of rated items (for rating-based by-user) or the set of touched items (for event-based by-user)
    - Return `candidates - exclude`

mdekstrand: Under the hood, `UserBasedCandidateIdentifier` can delegate to `ItemBasedCandidateIdentifier`.

lenskit-cli/src/main/java/org/lenskit/cli/commands/GlobalRecommend.java
mdekstrand: I now know.

We should fold the code into `Recommend`, and have an option `--for-items` to recommend for items instead of users.

mdekstrand: Ehh, that makes some things super-weird.

We should come up with a better name, but perhaps that could be a separate ticket.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
mdekstrand: I have refiled this proposal under #797, so we can merge this code without changing the candidate selection mechanism and make that a separate change.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
mdekstrand: Fixed.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemScorer.java
mdekstrand: Fixed.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemScorer.java
mdekstrand: Fixed.

lenskit-cli/src/main/java/org/lenskit/cli/commands/GlobalRecommend.java
mdekstrand: Re-raised in #798.

lenskit-api/src/main/java/org/lenskit/api/ItemBasedItemRecommender.java
mdekstrand: Done.

lenskit-core/src/main/java/org/lenskit/util/collections/LongUtils.java
mdekstrand: Not yet.

lenskit-knn/src/main/java/org/lenskit/knn/item/ItemItemItemBasedItemScorer.java
mdekstrand: I think we should make this separation a separate change.

mdekstrand: Raised in #799.

mdekstrand:
I think I've addressed (almost) everything, either by code/doc changes or filing new issues.

kluver:
Sorry, missed your last comment (14 days ago) somehow, and didn't know this was still blocking on my (and continuous integration? that can't be right.)

kluver:
Yah, it looks good. Most of the tricky issues seem to have been raised in future enhancements (sounds like a good policy) and It looked like the documentation things I caught have been fixed. :+1: 

~Kluver

mdekstrand:
OK, merging. Thanks, @kluver.

The CI waiting is because Appveyor changed how they report CI status results, which is interacting entertainingly with our branch protection configuration.

PR #814, opened 2015-11-03
Use R + Renjin rather than Python to verify integration tests

mdekstrand:
This set of changes uses R (with Renjin, so it can run on just the JVM) instead of Python to verify integration tests.

This has a few limitations:

- we can't use many convenient R packages, including `dplyr` and `reshape2`, because they don't work on Renjin
- the R function `reshape` is also broken on Renjin, so we have to manually reshape data
- ~~we are currently pulling in Renjin over unauthenticated HTTP~~

The advantage of this, once it is done, is that Python will no longer be required in order to build and run (most of) LensKit's full test suite.


mdekstrand:
Attached is a prediction file that produces the Renjin error.
[predictions.txt](https://github.com/lenskit/lenskit/files/26490/predictions.txt)

akbertram:
Renjin's repository at nexus.bedatadriven.com is now available over https.
https://nexus.bedatadriven.com/content/groups/public/org/renjin/

mdekstrand:
@akbertram Awesome, thanks!

mdekstrand:
We're now downloading Renjin over HTTPS.

We still require Python for one test: the external algorithm test.

Issue #808, opened 2015-11-02
Support header lines in TextDataSourceSpec


mdekstrand:
The text data source spec does not currently support `headerLines`. What we need to do:

- [x] Modify `TextDataSourceSpec` to have a `headerLines` field, getter, and setter.
- [x] Modify `TextDataSource.fromSpec` to read this property.
- [x] Modify `TextDataSource.toSpec()` to write this property

mdekstrand:
This is finished in #817.

PR #817, opened 2015-11-06
update changes to enable header Lines

amit93:
This addresses #808.


mdekstrand:
Thanks! I have some comments inline (or will shortly).

Could you please restore the `.idea` files that were deleted?

lenskit-specs/src/main/java/org/lenskit/specs/data/TextDataSourceSpec.java
```diff
@@ -94,7 +111,7 @@ public void setBuilderType(String type) {
      * @param fields The list of fields.  Can be `null` to use the builder's default fields.
      */
     public void setFields(@Nullable List<String> fields) {
-        this.fields = fields != null ? new ArrayList<>(fields) : null;
+        this.fields = fields != null ? new ArrayList<String>(fields) : null;
```
mdekstrand: This change is not needed. Make sure that IntelliJ is set up to use the Java 7 language level in your Project Structure.

lenskit-specs/src/main/java/org/lenskit/specs/data/TextDataSourceSpec.java
```diff
@@ -123,7 +140,7 @@ public void setDomain(PrefDomainSpec domain) {
 
     @Override
     public Set<Path> getInputFiles() {
-        Set<Path> files = new HashSet<>();
+        Set<Path> files = new HashSet<Path>();
```
mdekstrand: Same for this change.

lenskit-test/src/main/java/org/grouplens/lenskit/util/test/ExtraMatchers.java
```diff
@@ -61,6 +61,6 @@ private ExtraMatchers() {}
      * @return A matcher that accepts objects equivalent to `obj`.
      */
     public static <T> Matcher<T> equivalentTo(T obj, Equivalence<T> equiv) {
-        return new EquivalenceMatcher<>(obj, equiv);
+        return new EquivalenceMatcher<T>(obj, equiv);
```
mdekstrand: And this change.

lenskit-core/src/main/java/org/grouplens/lenskit/data/source/TextDataSource.java
```diff
@@ -181,6 +182,7 @@ public static TextDataSource fromSpec(TextDataSourceSpec spec) {
         bld.setFormat(fmt);
         bld.setItemFile(spec.getItemFile());
         bld.setItemNameFile(spec.getItemNameFile());
+        bld.setHeaderLines(spec.getHeaderLines());
```
mdekstrand: This needs to be specified on the format, not on the builder. So it would be `fmt.setHeaderLines`. That will make the tests pass.

We could make an argument that `setFormat` should allow this to work. But that is a separate change, I think.

lenskit-specs/src/main/java/org/lenskit/specs/data/TextDataSourceSpec.java
```diff
@@ -48,6 +49,22 @@ public String getName() {
         }
     }
 
+    /**
+     * Get the header Line.
+     * @return The header Line.
+     */
```
mdekstrand: Let's describe what this is in the comment. 'Get the number of header lines to skip', '@return The number of header lines to skip.'

We should also say '@see DelimitedColumnEventFormat.getHeaderLines()'.

lenskit-specs/src/main/java/org/lenskit/specs/data/TextDataSourceSpec.java
```diff
@@ -48,6 +49,22 @@ public String getName() {
...
+    public int getHeaderLines() {
+        return headerLines;
+    }
+
+    /**
+     * Set the headerLines.
+     * @param headerLines
+     */
```
mdekstrand: Let us make similar changes to this comment.

mdekstrand:
@amit93 I've gone through, and found the problem that is causing the tests to fail.

lenskit-core/src/main/java/org/grouplens/lenskit/data/source/TextDataSource.java
```diff
@@ -181,6 +182,7 @@ public static TextDataSource fromSpec(TextDataSourceSpec spec) {
         bld.setFormat(fmt);
         bld.setItemFile(spec.getItemFile());
         bld.setItemNameFile(spec.getItemNameFile());
+        bld.setHeaderLines(fmt.getHeaderLines());
```
mdekstrand: Not quite.  This should be:

    fmt.setHeaderLines(spec.getHeaderLines());

And should probably be higher up, where you're setting up the format.

PR #819, opened 2015-11-10
Optimize dot products

mdekstrand:
Check vector length, and iterate over the shorter verctor.


mdekstrand:
Working on doing some micro-benchmarks to see how to really optimize these things. Right now, our dot products are pretty terrible compared to the old sparse vector.

mdekstrand:
See some benchmarking output: [human.txt](https://github.com/lenskit/lenskit/files/32312/human.txt)

mdekstrand:
The new sorted dot product logic is much closer to the old sparse vector logic in its performance.

Issue #827, opened 2015-11-11
Add a means to bundle up LensKit for artifact submission


mdekstrand:
For reproducible research, it would be good to provide a means to bundle up the particular version of LensKit used, along with all its dependencies, to submit them to an archive.

This could take the form of a local Maven repository that contains LensKit and its dependencies, along with their source JARs; the experiment can then just reference this repository.

mdekstrand:
This is turning out to be quite difficult. I can create a Maven repository containing LensKit easily enough, but despite online pointers that the artifact resolution API would help me do this, I have been unable to effectively get it working.

mdekstrand:
I think this is mostly working as of #831, but needs some additional testing.

mdekstrand:
Have this working and tested.

PR #831, opened 2015-11-19
Implement repo bundling

mdekstrand:
This implements repo bundling to address #827.

Would like to have the source artifacts included as well.


mdekstrand:
Dropping in favor of the #832 solution.

Issue #835, opened 2015-11-23
Support isolation on crossfolding


mdekstrand:
We need to re-enable isolation groups on the crossfolder and get them routed through the data sets.

mdekstrand:
This works now, with #836. Isolation is a feature of adding the data sets, not of generating them.

PR #836, opened 2015-11-23
Expose data set isolation in Gradle plugin

mdekstrand:
This addresses #835, exposing data set isolation in the Gradle plugin.

The chief means of accessing it is at the train-test level, applying an `isolate` option when you add a data set:

```groovy
task evaluate(type: TrainTest) {
    /* ... */
    dataSet crossfold, isolate: true
    /* ... */
}
```


mdekstrand:
I'm going to double-down on the `isolate` option, since there isn't a good way to route it all the way from the crossfolder.

Issue #838, opened 2015-11-26
An exception when test and train folds are disjoint


jumutc:
Hello, 
When I execute ``TrainTestExperiment`` with ``PredictEvalTask`` I get an exception when my rating matrix is small and cross-validation leads to the disjoint test and train folds:
```
org.lenskit.eval.traintest.EvaluationException: error running evaluation
	at org.lenskit.eval.traintest.TrainTestExperiment.runJobList(TrainTestExperiment.java:543)
	at org.lenskit.eval.traintest.TrainTestExperiment.execute(TrainTestExperiment.java:382)
	at ...
Caused by: java.lang.NullPointerException: null
	at org.lenskit.eval.traintest.predict.PredictEvalTask$PredictConditionEvaluator.finish(PredictEvalTask.java:301)
	at org.lenskit.eval.traintest.ExperimentJob.run(ExperimentJob.java:181)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	... 1 common frames omitted
```

I have fixed it with a small modification in the ``finish`` method of ```PredictEvalTask``` but it would be nice to have it in the ``lenskit`` framework codebase as well.

mdekstrand:
Thanks for filing this!

I can take a look at it, but if you'd like to submit your fix as a pull request, I'd gladly consider integrating it directly.

Could you say a bit more about exactly how your train/test folds are disjoint? Are there users in testing that aren't in training, or items?

jumutc:
Thank you for a quick response! I will probably submit my pull request because for my project I had to reimplement ``PredictEvalTask`` to fix the problem. You are indeed right -> in test there are users which are not present in train but ``java.lang.NullPointerException`` occurs only when entire test set is unseen by the trained model.

jumutc:
Thank you for merging!

PR #839, opened 2015-11-27
fixing PredictEvalTask for disjoint train and test sets

jumutc:
Please take a look at my fix for issue #838

### DIFF
--- a/lenskit-eval/build.gradle
+++ b/lenskit-eval/build.gradle
@@ -39,6 +39,7 @@ apply from: "$rootDir/gradle/test-utils.gradle"
 
 dependencies {
     compile project(':lenskit-core')
+    compile project(':lenskit-knn')
     compile project(':lenskit-groovy')
     compile group: 'joda-time', name: 'joda-time', version: '2.3'
     compile group: 'org.apache.ant', name: 'ant', version: '1.8.4'
--- a/lenskit-eval/src/main/java/org/lenskit/eval/traintest/predict/PredictEvalTask.java
+++ b/lenskit-eval/src/main/java/org/lenskit/eval/traintest/predict/PredictEvalTask.java
@@ -298,7 +298,15 @@ public class PredictEvalTask implements EvalTask {
         public Map<String, Object> finish() {
             Map<String,Object> results = new HashMap<>();
             for (MetricContext<?> mc: predictMetricContexts) {
-                results.putAll(mc.getAggregateMeasurements().getValues());
+                MetricResult measurements = mc.getAggregateMeasurements();
+                // if test and train tests are disjoint we can get "null"
+                if (measurements != null) {
+                    results.putAll(measurements.getValues());
+                } else {
+                    for (String label : mc.metric.getAggregateColumnLabels()) {
+                        results.put(label, Double.NaN);
+                    }
+                }
             }
             return results;
         }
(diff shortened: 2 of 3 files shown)


mdekstrand:
Great, thanks! I also very much appreciate having tests for the problem. I'll make a couple specific change requests in-line, and then we should be good to merge.

lenskit-eval/build.gradle
```diff
@@ -39,6 +39,7 @@ apply from: "$rootDir/gradle/test-utils.gradle"
 
 dependencies {
     compile project(':lenskit-core')
+    compile project(':lenskit-knn')
```
mdekstrand: I'd rather not have the eval module grow a dependency on kNN. Fortunately, there are two easy ways of fixing this:

- Change the `knn` dependency from `compile` to `testCompile` (so it's only pulled in for the tests)
- Alter the tests to use a baseline scorer instead of a k-NN scorer (e.g. `ItemMeanRatingItemScorer`)

Either one would be acceptable.

lenskit-eval/src/main/java/org/lenskit/eval/traintest/predict/PredictEvalTask.java
```diff
@@ -298,7 +298,15 @@ public PredictConditionEvaluator(TableWriter tw, RatingPredictor pred, List<Metr
         public Map<String, Object> finish() {
             Map<String,Object> results = new HashMap<>();
             for (MetricContext<?> mc: predictMetricContexts) {
-                results.putAll(mc.getAggregateMeasurements().getValues());
+                MetricResult measurements = mc.getAggregateMeasurements();
+                // if test and train tests are disjoint we can get "null"
+                if (measurements != null) {
+                    results.putAll(measurements.getValues());
+                } else {
```
mdekstrand: I don't think we need the `else` block - if we don't have it, then the missing values will be `null`, which is perfectly acceptable (they'll be emitted as empty cells in the final CSV file). Empty cells work better than NaN, because R will pick them up as NA (which is distinct from NaN in R's world).

lenskit-eval/build.gradle
jumutc: The problem is that exception is reproducible only with this particular scorer. So the change to ``testCompile`` dependency would be sufficient!

lenskit-eval/src/main/java/org/lenskit/eval/traintest/predict/PredictEvalTask.java
jumutc: The fix is exactly in these lines. You can revert my changes in ``PredictEvalTask`` and see how test is failing!

lenskit-eval/build.gradle
mdekstrand: Got it. I think I see why — baselines will still predict for unknown users. This error likely arises when only when no users could be predicted.

lenskit-eval/src/main/java/org/lenskit/eval/traintest/predict/PredictEvalTask.java
mdekstrand: I'm just talking about the `else` block, not the `if` itself. It's fine for `results` to be missing some values; the NPE is arising when dereferencing `measurements`.

Now, it turns out that `measurements` should never be null under any circumstances, which means one of our metrics has a bug. But this test is good for safety protection against bad metrics.

mdekstrand:
Merged now, with the root cause also fixed. Thanks, @jumutc.

Issue #840, opened 2015-11-27
Live-lock in evaluator


mdekstrand:
In certain rare cases, we encounter the live lock described in [JAVASERVERFACES-2544](https://java.net/jira/browse/JAVASERVERFACES-2544) in the evaluator.

Example stack trace:

```
"pool-1-thread-40" #61 prio=5 os_prio=0 tid=0x00007f1654a76000 nid=0x15825 runnable [0x00007eeee41eb000]
   java.lang.Thread.State: RUNNABLE
        at java.util.WeakHashMap.transfer(WeakHashMap.java:528)
        at java.util.WeakHashMap.resize(WeakHashMap.java:493)
        at java.util.WeakHashMap.put(WeakHashMap.java:466)
        at org.grouplens.grapht.InjectionContainer.makeInstantiator(InjectionContainer.java:118)
        at org.grouplens.grapht.InjectionContainer$DepLookup.apply(InjectionContainer.java:163)
        at org.grouplens.grapht.InjectionContainer$DepLookup.apply(InjectionContainer.java:143)
        at com.google.common.collect.Maps$AsMapView.get(Maps.java:793)
        at org.grouplens.grapht.reflect.internal.ClassInstantiator.instantiate(ClassInstantiator.java:93)
        at org.grouplens.grapht.Instantiators$ProviderInstantiator.instantiate(Instantiators.java:148)
        at org.grouplens.grapht.Instantiators$MemoizingInstantiator.instantiate(Instantiators.java:184)
        - locked <0x00007ef3c33aacb0> (a org.grouplens.grapht.Instantiators$MemoizingInstantiator)
        at org.lenskit.inject.NodeInstantiator$DefaultImpl.instantiate(NodeInstantiator.java:76)
        at org.lenskit.eval.traintest.ComponentCache$CacheEntry.getObject(ComponentCache.java:207)
        - locked <0x00007ef3c33aad50> (a org.lenskit.eval.traintest.ComponentCache$CacheEntry)
        at org.lenskit.eval.traintest.ComponentCache.processNode(ComponentCache.java:132)
        at org.lenskit.inject.NodeProcessors.processNodes(NodeProcessors.java:89)
        at org.lenskit.eval.traintest.ExperimentJob.buildRecommender(ExperimentJob.java:219)
        at org.lenskit.eval.traintest.ExperimentJob.run(ExperimentJob.java:98)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[...]

PR #841, opened 2015-11-27
Update to Grapht lifecycles

mdekstrand:
This updates LensKit to be binary-compatible with the lifecycle-based Grapht.

Still need to make use of lifecycles in LensKit.


mdekstrand:
For the record, this worked on #840.

Issue #751, opened 2015-07-09
Add jvmArgs configuration to gradle plugin


mdekstrand:
We want to support easy configuration of JVM arguments in the Gradle plugin, for running the LensKit evaluator.

There are several pieces that need to be done to implement this:

- [x] Add a `jvmArgs` property of type `List<String>` to `LenskitExtension`
- [x] Add a `jvmArgs(String...)` method to `LenskitExtension` that adds its arguments to the `jvmArgs` property
- [x] Modify the code in `LenskitPlugin` that sets extension properties from Gradle project properties so that if the type of a property (on the extension) is `List`, it splits the property value using `StrTokenizer` (from commons-lang), accepting whitespace as the delimiter and `"` and `'` as the quote characters. This can be done by creating a tokenizer like this:

    ```java
    StrTokenizer tok = new StrTokenizer(val, StrMatcher.splitMatcher(), StrMatcher.quoteMatcher());
    ```

- [ ] Modify `LenskitTask` to have the following:

    - A  `jvmArgs` property, of type `List<String>`
    - Code in the constructor to set a convention mapping so that `jvmArgs` defaults to `jvmArgs` from the extension
    - Code in `applyFinalSettings` to add all the arguments in `getJvmArgs()` to the `invoker.jvmArgs`

- [ ] Write a test to test this logic. It should be in `lenskit-integration-tests/src/it/gradle`. It should have a `build.gradle` file that sets up an evaluation, an `eval.groovy` that just checks for some test property (`test.prop`) in `System.getProperties()` to have the correct value (`test-value`), and a `gradle.properties` file that contains `lenskit.jvmArgs=-Dtest.prop=test-value`.

mdekstrand:
This is all done and working in #872.

PR #857, opened 2016-01-30
master

amit93:


lenskit-gradle/src/main/groovy/org/lenskit/gradle/LenskitExtension.groovy
```diff
@@ -72,4 +72,14 @@ public class LenskitExtension {
      * and the log file.
      */
     def String logFileLevel
+
+    /*
+* The list of jvm argument prroperties
+* */
+    def List<String> jvmArgs
+
+    def addJVMArgs(String val){
```
mdekstrand: A few changes we need here:

- This should take an array of strings (`String... args`).
- It should just be called `jvmArgs`.
- It needs to add the arguments to the `jvmArgs` list, not just return them.

mdekstrand:
This works on #751.

lenskit-gradle/src/main/groovy/org/lenskit/gradle/LenskitPlugin.groovy
```diff
@@ -51,6 +53,13 @@ public class LenskitPlugin implements Plugin<Project> {
                 if (prop.type != String) {
                     val = prop.type.metaClass.invokeConstructor(val)
                 }
+
+                // if the type is list update the val using strtokenizer
+                if (prop.type == List) {
```
mdekstrand: This should be integrated with the previous `if`, probably before it, so we have:

```java
if (prop.type == List) {
    // setup code
} else if (prop.type != String) {
    // original code
}
```

lenskit-gradle/src/main/groovy/org/lenskit/gradle/LenskitTask.groovy
```diff
@@ -72,9 +72,16 @@ public abstract class LenskitTask extends ConventionTask {
...
+    /*
+    * The list of jvm argument prroperties
+    * */
+    def List<String> jvmArgs
+
     LenskitTask() {
         invoker = new JavaExecHandleBuilder(services.get(FileResolver))
         def ext = project.extensions.getByType(LenskitExtension)
+        // map jvmargs default to the jvmargs from the extension
+        jvmArgs = { ext.jvmArgs }
```
mdekstrand: This needs to be `conventionMapping.jvmArgs`

lenskit-gradle/src/main/groovy/org/lenskit/gradle/LenskitTask.groovy
```diff
@@ -95,6 +102,10 @@ public abstract class LenskitTask extends ConventionTask {
         if (logbackConfiguration) {
             invoker.systemProperties 'logback.configurationFile': project.file(logbackConfiguration)
         }
+
+        // add all the arguments in the invoker
+        invoker.jvmArgs = getJvmArgs()
```
mdekstrand: Let's add this argument to the `jvmArgs` rather than replacing them:

```java
invoker.jvmArgs.addAll(getJvmArgs());
```

mdekstrand:
Thanks for putting this together, @amit93! I've made a few comments on some things that need to be changed. Also, we need to get the build working and the merge conflicts resolved.

lenskit-gradle/src/main/groovy/org/lenskit/gradle/LenskitPlugin.groovy
```diff
@@ -48,9 +50,15 @@ public class LenskitPlugin implements Plugin<Project> {
             if (project.hasProperty(prjProp)) {
                 def val = project.getProperty(prjProp)
                 logger.info 'setting property {} to {}', prjProp, val
+                // if the type is list update the val using strtokenizer
+                if (prop.type == List) {
+                    StrTokenizer tok = new StrTokenizer(val, StrMatcher.splitMatcher, StrMatcher.quoteMatcher());
+                    val = prop.type.metaClass.invokeConstructor(tok)
+                }
                 if (prop.type != String) {
```
mdekstrand: I think this should be an `else if`.

lenskit-gradle/src/main/groovy/org/lenskit/gradle/LenskitTask.groovy
```diff
@@ -95,6 +102,9 @@ public abstract class LenskitTask extends ConventionTask {
         if (logbackConfiguration) {
             invoker.systemProperties 'logback.configurationFile': project.file(logbackConfiguration)
         }
+
+        // add all the arguments in the invoker
+        invoker.jvmArgs.addAll(getJvmArgs())
```
mdekstrand: Good!

lenskit-gradle/src/main/groovy/org/lenskit/gradle/LenskitExtension.groovy
```diff
@@ -72,4 +72,15 @@ public class LenskitExtension {
...
      */
     def String logFileLevel
+
+    /*
+* The list of jvm argument prroperties
+* */
+    def List<String> jvmArgs
+
+    def jvmArgs(String... val){
+        jvmArgs.add(val)
```
mdekstrand: This should be `addAll` rather than `add`.

mdekstrand:
@amit93 looking better! We still need to get these merge conflicts sorted out, though.

mdekstrand:
We are re-doing this pull request from another branch.

PR #862, opened 2016-02-16
Cross-port popularity blending from 3.0-T

mdekstrand:
This ports #860, the popular item recommender and rank-blending hybrid, to LensKit Master.


lenskit-core/src/main/java/org/lenskit/hybrid/Left.java
```diff
@@ -0,0 +1,38 @@
...
+
+import javax.inject.Qualifier;
+import java.lang.annotation.*;
+
+/**
+ * The 'left' recommender for a hybrid.
+ */
+@Qualifier
+@Documented
+@DefaultImplementation(TopNItemRecommender.class)
```
mdekstrand: This causes a problem if we want to use `@Left` in an item scorer.

Proposed solutions:

- add a 'target' class type to `@DefaultImplementation`, so that we can say ‘when satisfying an `ItemScorer`, use a `TopNItemScorer`.
- support defaults on injection points, which we want to do anyway, and put it there.

mdekstrand: Another potential solution: allowing defaults as per grouplens/grapht#135, and removing these lines.

kluver: Generally I feel that this is part of the tensions between reusing annotations (good in some places) and having annotations be specific to one class (good in other places). Is it possible to have an annotation be a subclass of another class so it would be `@RankBlendingItemScorer.Left`? That would avoid the concern in this case where "left" is too generic to be a reused annotation

mdekstrand: @kluver That would be possible. Not sure if I like it :) but it is possible.

PR #863, opened 2016-02-17
Useless parentheses around expressions should be removed to prevent any misunderstanding

AymanDF:
This pull request is focused on resolving occurrences of Sonar rule squid:UselessParenthesesCheck - “ Useless parentheses around expressions should be removed to prevent any misunderstanding ”. You can find more information about the issue here:
https://dev.eclipse.org/sonar/rules/show/squid:UselessParenthesesCheck
Please let me know if you have any questions.
Ayman Abdelghany.

### DIFF
--- a/lenskit-cli/src/main/java/org/lenskit/cli/util/InputData.java
+++ b/lenskit-cli/src/main/java/org/lenskit/cli/util/InputData.java
@@ -126,7 +126,7 @@ public class InputData {
     @Nullable
     public EventDAO getEventDAO() throws IOException {
         DataSource src = getSource();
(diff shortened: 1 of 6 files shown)


mdekstrand:
@AymanDF Thanks for this!

Most of the changes are in ternary operators, where I tend to prefer to the extra parentheses to make it easier to read the expression. If there's a good counterargument, though, I'm happy to hear it!

But I think that I'd be likely to disable this particular rule in our Sonar instance, as extra parentheses can often aid human understanding. Sometimes they don't — the `assert` changes in this PR are a great example of where this rule is useful — but many times they do.

PR #884, opened 2016-03-16
Add NDPM metric

vaibhav345:
Implementation of NDPM Metric and getUnseenItem() method, and some small editing issues solved.


mdekstrand:
Compile error in the NDPM metric.

vaibhav345:
I have fixed the error.

lenskit-eval/src/main/java/org/lenskit/eval/traintest/metrics/MapMetricResult.java
```diff
@@ -32,9 +32,7 @@
 class MapMetricResult extends MetricResult {
     private final Map<String, Object> values;
 
-    public MapMetricResult(Map<String,?> vals) {
-        values = new LinkedHashMap<>(vals);
-    }
+    public MapMetricResult(Map<String,?> vals) {values = new LinkedHashMap<>(vals);}
```
mdekstrand: In the future, please be careful to avoid spurious code changes like this.

lenskit-eval/src/main/java/org/lenskit/eval/traintest/recommend/TopNNDPMMetric.java
```diff
@@ -0,0 +1,139 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Collections;
+
+/**
+ * Measure the nDPM of the top-N recommendations, using rankings.
+ * This metric is registered with the type name `ndpm`.
+ * The paper used as a reference for this implementation is http://www2.cs.uregina.ca/~yyao/PAPERS/jasis_ndpm.pdf.
+ * Created by VaibhavMahant on 2/1/16.
```
mdekstrand: We do not generally sign our code files — the relevant information is in the Git committer log.

lenskit-eval/src/main/java/org/lenskit/eval/traintest/recommend/TopNNDPMMetric.java
```diff
@@ -0,0 +1,139 @@
...
+            return MetricResult.empty();
+        }
+
+        Long2DoubleMap ratings = user.getTestRatings();
+
+        long[] actual = LongUtils.asLongCollection(recommendations.idList()).toLongArray();
+
+        double dpm = computeDPM(actual, ratings);
+
+        double nDPM = dpm / 2 * dpm; // Normalized nDPM
```
mdekstrand: Is this correct? My read of the paper is that the denominator should be 2C. Also, when dividing by a product, we must put the denominator in parentheses.

lenskit-eval/src/main/java/org/lenskit/eval/traintest/recommend/TopNNDPMMetric.java
```diff
@@ -0,0 +1,139 @@
...
+                if (value.containsKey(actual_item[i])) {
+                    valueOne = value.get(actual_item[i]);
+
+                    if (value.containsKey(actual_item[j])) {
+                        valueTwo = value.get(actual_item[j]);
+                    }
+                }
+                if(valueOne < valueTwo)
+                    nCompatible++;
+                if(valueOne == valueTwo)
```
mdekstrand: what if ratings are actually 0? Then it will say they disagree, when they do not, won't it?

mdekstrand:
@vaibhav345 I've made a couple comments on logic issues. I'm not convinced the math is correct yet.

lenskit-eval/src/main/java/org/lenskit/eval/traintest/recommend/TopNNDPMMetric.java
```diff
@@ -0,0 +1,175 @@
...
+                    if (value.containsKey(actual_item[j])) {
+                        valueTwo = value.get(actual_item[j]);
+                    }
+                }
+                if(valueOne < valueTwo){
+                    nDisagree++;
+                }
+                if(valueOne == valueTwo) {
+                    nCompatible++;
+                }
```
mdekstrand: Is it correct to count items as compatible if the user's ordering does not assert anything about them?

vaibhav345: Yes, we need to count the number of item pair whose rating values are same.

mdekstrand: @vaibhav345 What about item pairs for which one of the items has no rating value?

Issue #845, opened 2015-12-07
Make generalized key-value interfaces for data


mdekstrand:
As per the discussion in the mailing list, we want to have a new key-value conception of data in LensKit.

### Work to Do

- [x] Entity interface (#917)
- [x] DAO interface (#914)
- [x] File-based DAOs (#914)
- [x] Implement specialized entity types (#923)
- [x] Implement bridge DAOs (#925)
- [x] Support crossfolding new DAOs (#929)
- [x] Use new DAOs in evaluator (#930)
- [x] Use new DAOs in command-line tools (#925)
- [x] Support preference domains in new DAOs (#931)

### Background

The idea of the event setup:

- Event is an interface that exposes a key-value interface, with unboxed calls for keys storing longs or doubles (I don’t think we need to worry about other unboxed primitives), and designated getters for fetching user ID, item ID, and timestamp.
- Rating implements Event, and exposes its rating under the ‘rating’ key.
- There’s a general event implementation backed by a hashtable.
- Additional data sources can provide their own implementations as appropriate.
- We may or may not have event IDs.
- Either vector DAOs or summarizing components are available to translate to vectors in item-space (or latent-feature-space, or whatever)

And how we can do DAOs:

> It seems that we need just a few concepts:
>
> - Object types. Most data stores will have user, item, and some kind of event such as ‘rating’.
> - Object IDs. IDs are namespaced per-type, so it is allowed for there to exist a user and item with the same ID as far as LensKit is concerned (However, LensKit never creates objects, so it is fine for the underlying data store to require and/or force all IDs to be disjoint). I think I’d like to keep IDs being longs. This does mean that events need to have IDs again.
> - Objects themselves, represented as key-value dictionaries optionally implemented by an event-specific class.
>
> The DAO needs to support the following operations:
> 
> - Fetch the IDs for a particular object type.
> - Stream objects of a particular type
> - Look up objects by certain common keys (e.g. user ID, item ID)
> - Look up objects by arbitrary key (optional operation, can be optimized by the DAO if desired)
>
> The last operation can be used for supporting non-long IDs, by allowing items or users to be looked up by string or UUID or whatever.

mdekstrand:
ping @kluver so he can track status on this ticket.

PR #914, opened 2016-06-14
Implement new DAO interface & static file DAOs

mdekstrand:
This works on #845, implementing new DAOs around the entities (#917).


lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
+# Data Manifest Schema
+
+This document describes the format of data manifests.
+
+Data manifests are JSON-style data.  Currently they are generally described in YAML files, but they may be described in other formats such as JSON.
```
kluver: so data manifests are yaml style data, which can be in other formats like json?

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Overall Layout
+
+A manifest file represents either a list or a map.
```
kluver: given that you are defining things in terms of json style data types it might be worth making a stronger claim above.

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+	columns:
+	    movieId: id
+	    title: name
+```
+
+## Data Source Description
+
+Individual data sources are described with the following schema.
+
+`type`
```
kluver: type = textfile is default?

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+:   The data source type.  Currently only `textfile` is supported.
+
+The remainder of the keys are defined by the particular data source.
+
+### `textfile` data sources
+
+`textfile` sources read data from text files, usually line-delimited.
```
kluver: not clear what ussually line-delimited means here. In particular I think this can only support simple delineated files right? Like I don't even know if we support the more exotic flavors of csv (strings enclosing fields with commas and new lines in them, the commas and newlines should be ignored)

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+:   The file format.  Can be one of:
+	 - `delimited` — delimited, columnar text (default delimiter is `\t`)
+	 - `csv` — `delimited` with a delimiter of `,`
+	 - `tsv` — tab-separated (`delimited` with delimiter of `\t`)
+
+`entity_type`
+:   The name of the entity type contained in this file.  The entity type is also used to provide defaults for the columns.  The default is `rating`.
+
+`builder`
+:   The entity builder to be used for these entities.  The entity type may provide a default; otherwise, `org.lenskit.data.entities.BasicEntityBuilder` is used.  The keyword `basic` can be used to refer to the basic entity builder, to override a default entity builder if desired.
```
kluver: magic keywords (`basic`) mixed with full class names is a strange mix to me.

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+
+`delimiter`
```
kluver: This should be documented right next to the delimited format.

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+:   The delimiter string for `delimited` files.
```
kluver: `Teh delimiter string for files with the \`delimited\` format

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+
+`header`
+:   Whether the file has a header.  If `true`, the file has a single-line header; if `false`, no header is assumed.  If an integer, it is the number of lines to skip (the lines are *not* interpreted as containing field names).  The default is `false`.
+
+`columns`
+:   A map or list describing the columns in the file (for columnar formats).  If `header` is `true`, then this can be a map whose keys are column header labels and whose values are column descriptors; otherwise, it is a list of column descriptors.  If any column is missing or has a `null` descriptor, then that column is ignored.
+
+	A column descriptor can be either a string, giving a column name, or a map with keys `name` (the column name) and `type` (the column type).
```
kluver: so the reader assumes that the csv has a header row and the header column tells it how many rows to ignore to get to the header row?

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+	
+`indexes`
```
kluver: yah people reading this document won't be a headspace to understand the signfiicance / use of this field

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+:   A list of attribute names to be indexed for fast lookup.  If no indexes are specified, item and user IDs are indexed by default if present.
+	
+`meta`
+:   Metadata about the data, such as the `domain` for rating values.`
```
kluver: so meta should be a dictionary and contains use-specific information?

this use-specific information is documented where?

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+:   Java `long`.
+
+`double`, `real`, or `Double`
+:   Java `double`.
+
+`string` or `String`
+:   Java `String`.
+
+Java class name
+:   The corresponding class.  Must be convertible with Joda-Convert.
```
kluver: link to a document explaining joda convert. feel free to link out of lenskit

lenskit-core/specs/data-manifest.md
```diff
@@ -0,0 +1,111 @@
...
+:   A list of attribute names to be indexed for fast lookup.  If no indexes are specified, item and user IDs are indexed by default if present.
+	
+`meta`
+:   Metadata about the data, such as the `domain` for rating values.`
+	
+## Attribute Data Types
+
+The following types are supported for attributes:
+
+`int` or `Integer`
```
kluver: maybe this is a yaml think I don't know, but where is type specified in all this? unless this is supposed to be up by the type field eariler.

lenskit-core/src/main/java/org/grouplens/lenskit/data/sql/JDBCRatingDAO.java
```diff
@@ -136,8 +136,14 @@ public JDBCRatingDAO(Connection dbc, SQLStatementFactory sfac, boolean close) {
         userStatement = new CachedPreparedStatement(dbc, statementFactory.prepareUsers());
         itemStatement = new CachedPreparedStatement(dbc, statementFactory.prepareItems());
         for (SortOrder order : SortOrder.values()) {
-            eventStatements.put(order, new CachedPreparedStatement(dbc,
-                                                                   statementFactory.prepareEvents(order)));
+            switch (order) {
+            case ASCENDING:
```
kluver: I'm not going to claim I understand whats going on here, but this doesn't make sense in context, especially a break instead of a continue.

lenskit-core/specs/data-manifest.md
mdekstrand: It just means that there's one line per entity. I think we might want to just make that a hard requirement, and if you've got some other kind of data write your own DAO.

One line per file is sufficiently general to support JSON instead of CSV, again one per line.

lenskit-core/specs/data-manifest.md
mdekstrand: Yes.

lenskit-core/specs/data-manifest.md
mdekstrand: YAML is a superset of JSON, and we don't use any of its semantic extensions, only its syntactic ones. Is there a better way to discuss this?

lenskit-core/specs/data-manifest.md
mdekstrand: good point.

lenskit-core/specs/data-manifest.md
mdekstrand: If the `columns` data structure is a map, then yes, the CSV file must have a header. If you try to use a `map` and *don't* configure `header` to `true`, then it detects the case and yells at you.

lenskit-core/src/main/java/org/grouplens/lenskit/data/sql/JDBCRatingDAO.java
mdekstrand: oof, `break` is confusing in this case. It breaks out of the `switch`, which, since the `switch` is the last element of the `for`, continues the `for`.

This particular code is a temporary stop-gap while both DAO interfaces are in LensKit. As soon as the old DAOs are gone, this code will be gone.

lenskit-core/specs/data-manifest.md
kluver: Data manifests are written in standard yaml.

lenskit-core/specs/data-manifest.md
kluver: document that somewhere

lenskit-core/specs/data-manifest.md
kluver: would this be the correct data source to do json? or would you recommend making that its own, more flexible data source?

If we only want to use one-entry-per-line delineated data we should just call it delimiter separated data (https://en.wikipedia.org/wiki/Delimiter-separated_values)

lenskit-core/specs/data-manifest.md
kluver: that is complicated and interacts strangely with the header. I think this needs a local re-design. map only where header is true. none of this skip and then header garbage. if header is false and columns is there then list. if header is true or an int then list is acceptable and all header rows will be ignored.

lenskit-core/src/main/java/org/grouplens/lenskit/data/sql/JDBCRatingDAO.java
kluver: OK. just remember to fix it later since making ascending and descending the same case sounds problematic.

lenskit-core/specs/data-manifest.md
mdekstrand: This is just specifying valid values for the `type` field. Perhaps I should place an explicit cross-reference when defining `type`.

lenskit-core/src/main/java/org/lenskit/data/dao/AbstractDataAccessObject.java
```diff
@@ -0,0 +1,107 @@
...
+import javax.annotation.WillCloseWhenClosed;
+import java.util.List;
+
+/**
+ * Helper class to make it easier to create DAOs.  This base class implements several of the DAO
+ * convenience methods in terms of the master {@link #streamEntities(EntityQuery)} method, so that
+ * DAO implementers only need to translate {@link EntityQuery} objects into their underlying query
+ * framework (collection predicates, Jooq queries, HQL, or whatever).
+ */
+public abstract class AbstractDataAccessObject implements DataAccessObject {
```
kluver: TODO: return and review after reviewing DAO

lenskit-core/specs/data-manifest.md
kluver: Is this where you specify IDs? (like which row/column is the itemID?) or is this different?

lenskit-core/src/main/java/org/lenskit/data/dao/DataAccessObject.java
```diff
@@ -0,0 +1,108 @@
...
+ * LensKit data is represented by *entities*; see {@link org.lenskit.data.entities}
+ *
+ * DAOs **must** be thread-safe, but individual streams they return need not be.
+ */
+public interface DataAccessObject {
+    /**
+     * Get the types of entities that are available in this DAO.
+     * @return The set of available entity types.
+     */
+    Set<EntityType> getEntityTypes();
```
kluver: I might find an answer as I get to queries - but should one dao manage multiple data types or should we have our data loading facilities return multiple daos one for each data type?

lenskit-core/src/main/java/org/lenskit/data/dao/EntityCollectionDAO.java
```diff
@@ -0,0 +1,156 @@
...
+        List<Attribute<?>> filters = query.getFilterFields();
+        if (filters.isEmpty()) {
+            candidates = data;
+        } else {
+            // optimize by trying to look up the first condition
+            Attribute<?> f1 = filters.get(0);
+            candidates = data.find(f1);
+        }
+
+        // FIXME don't filter if we don't need to
```
kluver: is this fixme documented as an enhancement?

lenskit-core/src/main/java/org/lenskit/data/dao/EntityCollectionDAO.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        Collection<Entity> candidates;
+        List<Attribute<?>> filters = query.getFilterFields();
+        if (filters.isEmpty()) {
+            candidates = data;
+        } else {
+            // optimize by trying to look up the first condition
+            Attribute<?> f1 = filters.get(0);
+            candidates = data.find(f1);
```
kluver: I don't understand this yet.

lenskit-core/src/main/java/org/lenskit/data/dao/EntityCollectionDAOBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+     * so far, so it should be called *after* all other calls to {@link #addEntity(Entity)} and friends.  If an entity
+     * has already been added with the same type and ID as one of the derived entites, it is kept instead of the derived
+     * entity.
+     *
+     * @param derived The derived entity type.
+     * @param source The source entity type.
+     * @param attr The source attribute.
+     * @return The builder (for chaining).
+     */
+    public EntityCollectionDAOBuilder deriveEntities(EntityType derived, EntityType source, TypedName<Long> attr) {
```
kluver: to make sure I'm clear, this is for derinving items and users from ratings, things like that right?

lenskit-core/src/main/java/org/lenskit/data/dao/EntityCollectionDAO.java
kluver: Ohh we renamed things so that attribute means key and value didn't we?

lenskit-core/src/main/java/org/lenskit/data/dao/EntityQuery.java
```diff
@@ -0,0 +1,118 @@
...
+        return viewType;
+    }
+
+    /**
+     * Call {@link #matches(Entity)}.
+     * @param entity The entity to match.
+     * @return `true` if the entity matches the query
+     */
+    @Override
+    public boolean apply(@Nullable Entity entity) {
```
kluver: This seems like a strange problem decomposition. Especially given that matches is public and could be called with a null (it isn't annotated nullable, but it isn't annotated nonull either)

lenskit-core/src/main/java/org/lenskit/data/dao/EntityQuery.java
```diff
@@ -0,0 +1,118 @@
...
+        return entity != null && matches(entity);
+    }
+
+    /**
+     * Determine whether this query matches the specified entity.
+     * @param entity The entity to test.
+     * @return The entity.
+     */
+    public boolean matches(Entity entity) {
+        if (!entity.getType().equals(entityType)) {
```
kluver: equals or assignable?

lenskit-core/src/main/java/org/lenskit/data/dao/EntityQueryBuilder.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Add a filter condition to the builder.
+     * @param name The attribute to filter.
+     * @param obj The object to test for equality.
+     * @param <A> The attribute type.
+     * @return The query builder (for chaining).
+     */
+    public <A> EntityQueryBuilder addFilterField(TypedName<A> name, @Nonnull A obj) {
```
kluver: only allowing attr is exactly style filters?

lenskit-core/specs/data-manifest.md
mdekstrand: that is reasonable.

lenskit-core/src/main/java/org/lenskit/data/dao/EntityCollectionDAOBuilder.java
mdekstrand: Exactly.

lenskit-core/src/main/java/org/lenskit/data/dao/DataAccessObject.java
mdekstrand: Current plan is a mono-DAO, which is a change from the LensKit 2 design.

lenskit-core/src/main/java/org/lenskit/data/dao/EntityQueryBuilder.java
mdekstrand: For now, yes. I would like to add less/greater filters later.

lenskit-core/src/main/java/org/lenskit/data/dao/DataAccessObject.java
kluver: why?

lenskit-core/src/main/java/org/lenskit/data/dao/EntityQuery.java
mdekstrand: It's just there to allow queries to be used with Guava predicate-related utilities. I think `apply` is a confusing name for use in code that specifically uses a query, so `matches` is the ‘real’ one.

kluver: why different contracts RE: nulls?

lenskit-core/specs/data-manifest.md
mdekstrand: Basically. It's like SQL indexes - it tells the data system to generate optimized lookup tables for certain columns.

kluver: then, no this isn't where you specify your entities id. should specifing the entities ID be special?

lenskit-core/src/main/java/org/lenskit/data/dao/file/DelimitedColumnEntityFormat.java
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * Delimited text column entity format.
+ */
+public class DelimitedColumnEntityFormat implements EntityFormat {
+    private String delimiter = "\t";
+    private int headerLines;
+    private boolean readHeader;
+    private EntityType entityType = EntityType.forName("rating");
+    private Class<? extends EntityBuilder> entityBuilder = BasicEntityBuilder.class;
+    private Constructor<? extends EntityBuilder> entityBuilderCtor;
```
kluver: ... factoryfactory

lenskit-core/src/main/java/org/lenskit/data/dao/file/DelimitedColumnEntityFormat.java
```diff
@@ -0,0 +1,235 @@
...
+        if (entityBuilderCtor == null || !entityBuilderCtor.getDeclaringClass().equals(entityBuilder)) {
+            try {
+                entityBuilderCtor = entityBuilder.getConstructor(EntityType.class);
+            } catch (NoSuchMethodException e) {
+                throw new IllegalArgumentException("cannot find suitable constructor for " + entityBuilder);
+            }
+        }
+        try {
+            return entityBuilderCtor.newInstance(entityType);
+        } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
```
kluver: thats a neat syntax. havn't seen that before.

lenskit-core/src/main/java/org/lenskit/data/dao/file/DelimitedColumnEntityFormat.java
```diff
@@ -0,0 +1,235 @@
...
+            throw new RuntimeException("could not instantiate entity builder", e);
+        }
+    }
+
+    /**
+     * Add a column.
+     * @param attr The attribute to add as a column, or `null` to skip the next column.
+     * @throws IllegalStateException if non-labeled columns have already been specified.
+     */
+    public void addColumn(@Nullable TypedName<?> attr) {
```
kluver: `Label the next column. The first time this is called it specifies the label for the first column. The second time this is called it specifies the second column label and so forth` or something like that to make the ordering clear.

lenskit-core/src/main/java/org/lenskit/data/dao/file/DelimitedColumnEntityFormat.java
```diff
@@ -0,0 +1,235 @@
...
+            columns = new ArrayList<>();
+        }
+        columns.add(attr);
+    }
+
+    /**
+     * Add columns to a format.
+     * @param columns The columns to add.
+     */
+    public void addColumns(TypedName<?>... columns) {
```
kluver: I'm not sure if I want to allow this to be used if the other ordered column add is used... At the vary least it needs a verbose comment like the past one.

lenskit-core/src/main/java/org/lenskit/data/dao/file/DelimitedColumnEntityFormat.java
```diff
@@ -0,0 +1,235 @@
...
+            tokenizer = tok;
+        }
+
+        @Override
+        public Entity parse(String line) {
+            tokenizer.reset(line);
+            lineNo += 1;
+
+            EntityBuilder builder = newEntityBuilder()
+                    .setId(lineNo);
```
kluver: ... and the first time I was aware that a default id would be provided. I like this behavior (esp. for ratings which have a natural ID that is a union of other columns) but it should be documented.

lenskit-core/src/main/java/org/lenskit/data/dao/file/EntitySource.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Interface for entity providers.
+ */
+public interface EntitySource {
+    /**
+     * Get the data from this entity source.
+     * @return The data from the entity source.
+     */
+    ObjectStream<Entity> openStream() throws IOException;
+}
```
kluver: I don't know lenskit code style but it stands out to me that this is the first no ending newline that I have seen.

lenskit-core/src/main/java/org/lenskit/data/dao/file/Formats.java
```diff
@@ -0,0 +1,49 @@
...
+ * Utility methods relating to entity formats.
+ */
+public final class Formats {
+    /**
+     * Read a CSV file of ratings without a header.  The ratings are assumed to be in `user, item, rating[, timestamp]`
+     * format.
+     *
+     * @return An event format reading ratings without a header.
+     */
+    public static EntityFormat csvRatings() {
```
kluver: will it deal with missing timestamp reasonably?

lenskit-core/src/main/java/org/lenskit/data/dao/file/DelimitedColumnEntityFormat.java
```diff
@@ -0,0 +1,235 @@
...
+
+            // since ID is already set, a subsequent ID column will properly override
+
+            for (TypedName column: fileColumns) {
+                String value = tokenizer.nextToken();
+                if (value != null && column != null) {
+                    builder.setAttribute(column, column.parseString(value));
```
kluver: If I read this right, missing data = null not error?

lenskit-core/src/main/java/org/lenskit/data/dao/file/StaticFileDAOProvider.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * Get the data access object. This method is thread-safe.
+     * @return The access object.
+     */
+    @Override
+    public DataAccessObject get() {
+        SoftReference<DataAccessObject> cache = cachedDao;
+        DataAccessObject dao = cache != null ? cache.get() : null;
```
kluver: So If I am reading this right, cacheing occurs at an "all or nothing" level.

lenskit-core/src/main/java/org/lenskit/data/dao/file/StaticFileDAOProvider.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    private static void configureDataSource(StaticFileDAOProvider layout, String name, JsonNode object, URI base) {
+        if (name == null) {
+            name = object.path("name").asText("<unnamed>");
+        }
+
+        String type = object.path("type").asText("textfile").toLowerCase();
+        if (!type.equals("textfile")) {
+            // FIXME Delegate file formats
```
kluver: is this fixme logged?

lenskit-core/src/main/java/org/lenskit/data/dao/file/TextEntitySource.java
```diff
@@ -0,0 +1,284 @@
...
+        int headerLines = format.getHeaderLines();
+        List<String> header = new ArrayList<>();
+        while (header.size() < headerLines) {
+            String line = lines.readObject();
+            if (line == null) {
+                IOException ex = new IOException(String.format("expected %d header lines, found %d", headerLines, header.size()));
+                try {
+                    lines.close();
+                } catch (Throwable th) {
+                    ex.addSuppressed(th);
```
kluver: catching throwable scares me some exceptions (out of memory) need to fail pretty darn fast. I have trouble thinking about java behavior in severe fail conditions. is there really value in catching Errors?

lenskit-core/src/main/java/org/lenskit/data/dao/file/TextEntitySource.java
```diff
@@ -0,0 +1,284 @@
...
+
+        JsonNode ebNode = object.path("builder");
+        if (ebNode.isTextual()) {
+            String ebName = ebNode.asText();
+            if (ebName.equals("basic")) {
+                format.setEntityBuilder(BasicEntityBuilder.class);
+            } else {
+                Class bld;
+                try {
+                    // FIXME Use a configurable class loader
```
kluver: fixme

lenskit-core/src/main/java/org/lenskit/data/dao/EntityQuery.java
mdekstrand: `apply`'s nullability contract is broken (it's annotated `@Nullable`, so subclasses must also be so annotated, but its JavaDoc specifies that null-tolerance is optional). However, there are very few, if any, contexts in which it makes sense to apply a query to a null entity.

lenskit-core/src/main/java/org/lenskit/data/dao/file/DelimitedColumnEntityFormat.java
mdekstrand: Yes. A particular entity builder, such as `RatingBuilder`, is free to fail when it gets to `build()` — that is how I intend to enforce mandatory fields, rather than making field specification more complicated.

lenskit-core/src/main/java/org/lenskit/data/dao/file/StaticFileDAOProvider.java
mdekstrand: Yes, for now. Right now I am trying to get the public API and a basic, correct implementation down.

I would like to add significantly more optimization, decrease memory usage, etc. once we have that in place.

lenskit-core/src/main/java/org/lenskit/data/dao/file/TextEntitySource.java
mdekstrand: To my knowledge, this is exactly the code that Java would generate if we used `try`-with-resources. However, since in the non-error case we return an (unclosed) stream, we can't actually use `try`-with-resources.

kluver:
I would be lying if I said I fully grocked all of the code. I think this big/dense of a PR is a bit long to do all in one sitting. That said everything looked OK. theres a few documentation changes I would ask before merge.

mdekstrand:
@kluver Thanks! Yeah, it's big, and there's a bunch of *new* stuff which is (IMO) harder to review than changes to existing code. I'll get this doc stuff worked up before merge.

lenskit-core/src/main/java/org/lenskit/data/dao/DataAccessObject.java
mdekstrand: Two reasons:

1. Make it a lot easier to connect and disconnect DAOs without learning about strange concepts such as 'model retention' and Grapht rewriting. We'll use Grapht rewriting under the hood, but it'll be easier to say 'you put the whole DAO in, you take the whole DAO out, you put the whole DAO in, and you shake it all about'.

2. Implement a single class to connect to new data sets.

3. Multiple DAOs with different data types are a little hard to inject.

lenskit-core/src/main/java/org/lenskit/data/dao/EntityCollectionDAO.java
mdekstrand: no, i'm just fixing it.

lenskit-core/src/main/java/org/lenskit/data/dao/file/StaticFileDAOProvider.java
mdekstrand: #922.

mdekstrand:
Yay, all the builds pass!

lenskit-core/src/main/java/org/lenskit/data/dao/file/DelimitedColumnEntityFormat.java
mdekstrand: I have documented this in the data manifest docs.

lenskit-core/src/main/java/org/lenskit/data/dao/file/DelimitedColumnEntityFormat.java
mdekstrand: New in Java 7.

lenskit-core/src/main/java/org/lenskit/data/dao/DataAccessObject.java
kluver: I guess I'm OK with this. In this design. I could see this introducing some extra objects whose purpose is to be a data structure cacheing and optimizing over the DAO. An example could be a user rating count or item rating count object. In LK2 these types of "queries" could reasonably fall under the guise of a DAO, but in this model I think they become something else.

mdekstrand: Agreed, and there's a little bit of funny business with that if we want those to e.g. query from the database. But I think it's all solvable funny business.

PR #917, opened 2016-06-17
Introduce new Entity data type

mdekstrand:
This PR begins work on #845 by implementing the `Entity` data type that will form the new foundation of LensKit's data model.

Entities are identified with a *type* and an *id*; the ID must be unique for all entities of a particular type (think of it as a primary key).

Entities also have attributes with values. A type-safe API provides access to attribute values.

It's intended that most entity types have associated *view classes*, such as `Rating`, that provide more efficient storage and more type-safe access for code that knows the data type it's working with. The `Entity` interface provides a generic way of describing any data, and allows code to be data-type-independent.

This is a prerequisite for #914.

### DIFF
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
@@ -1,0 +1,177 @@
+/*
+ * LensKit, an open source recommender systems toolkit.
+ * Copyright 2010-2014 LensKit Contributors.  See CONTRIBUTORS.md.
+ * Work on LensKit has been funded by the National Science Foundation under
+ * grants IIS 05-34939, 08-08692, 08-12148, and 10-17697.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.data.entities;
+
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+
+import javax.annotation.Nonnull;
+import java.util.*;
+
+/**
+ * Base class to make it easier to implement entities.
+ */
+public abstract class AbstractEntity implements Entity {
+    @Override
+    public Map<String, Object> asMap() {
+        return new MapView();
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * This implementation delegates to {@link #maybeGet(Attribute)}.
+     */
+    @Nonnull
+    @Override
+    public <T> T get(Attribute<T> attribute) {
+        T val = maybeGet(attribute);
+        if (val == null) {
+            throw new NoSuchAttributeException(attribute.toString());
+        } else {
+            return val;
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * This implementation delegates to {@link #maybeGet(String)}.
+     */
+    @Nonnull
+    @Override
+    public Object get(String attr) {
+        Object val = maybeGet(attr);
+        if (val == null) {
+            throw new NoSuchAttributeException(attr);
+        } else {
+            return val;
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * This implementation delegates to {@link #get(Attribute)}.
+     */
+    @Override
+    public long getLong(Attribute<Long> attribute) {
+        return get(attribute);
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * This implementation delegates to {@link #get(Attribute)}.
+     */
+    @Override
+    public double getDouble(Attribute<Double> attr) {
+        return get(attr);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        } else if (obj instanceof Entity) {
+            Entity oe = (Entity) obj;
+            return getType().equals(oe.getType()) && getId() == oe.getId() && asMap().equals(oe.asMap());
+        } else {
+            return false;
+        }
+    }
+
+    @Override
+    public int hashCode() {
+        HashCodeBuilder hcb = new HashCodeBuilder();
+        hcb.append(getType())
+           .append(getId())
+           .append(asMap());
+        return hcb.toHashCode();
+    }
+
+    private class MapView extends AbstractMap<String,Object> {
+        @Override
+        public Set<Entry<String, Object>> entrySet() {
+            return new EntrySet();
+        }
+
+        @Override
+        public int size() {
+            return getAttributeNames().size();
+        }
+
+        @Override
+        public boolean containsKey(Object key) {
+            return key instanceof String && hasAttribute((String) key);
+        }
+
+        @Override
+        public Object get(Object key) {
+            if (key instanceof String) {
+                return maybeGet((String) key);
+            } else {
+                return null;
+            }
+        }
+
+        @Override
+        public Set<String> keySet() {
+            return getAttributeNames();
+        }
+    }
+
+    private class EntrySet extends AbstractSet<Map.Entry<String,Object>> {
+        @Override
+        public Iterator<Map.Entry<String, Object>> iterator() {
+            return new EntryIter(getAttributeNames().iterator());
+        }
+
+        @Override
+        public int size() {
+            return getAttributeNames().size();
+        }
+    }
+
+    private class EntryIter implements Iterator<Map.Entry<String,Object>> {
+        private final Iterator<String> baseIter;
+
+        public EntryIter(Iterator<String> keyIter) {
+            baseIter = keyIter;
+        }
+
+        @Override
+        public boolean hasNext() {
+            return baseIter.hasNext();
+        }
+
+        @Override
+        public Map.Entry<String, Object> next() {
+            String attr = baseIter.next();
+            return ImmutablePair.of(attr, get(attr));
+        }
+
+        @Override
+        public void remove() {
+            throw new UnsupportedOperationException();
+        }
+    }
+}
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
@@ -30,8 +30,6 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import javax.annotation.OverridingMethodsMustInvokeSuper;
-import javax.print.attribute.standard.MediaSize;
 import java.util.*;
 
 /**
@@ -59,13 +57,13 @@ public abstract class AbstractEntity implements Entity {
     /**
      * {@inheritDoc}
      *
-     * Delegates to {@link #getAttributeTypedNames()} and extracts the names.
+     * Delegates to {@link #getTypedAttributeNames()} and extracts the names.
      */
     @Override
     public Set<String> getAttributeNames() {
         // TODO Make this more efficient
         ImmutableSet.Builder<String> bld = ImmutableSet.builder();
-        for (TypedName<?> name: getAttributeTypedNames()) {
+        for (TypedName<?> name: getTypedAttributeNames()) {
             bld.add(name.getName());
         }
         return bld.build();
@@ -74,7 +72,7 @@ public abstract class AbstractEntity implements Entity {
     /**
      * {@inheritDoc}
      *
-     * Delegates to {@link #getAttributeTypedNames()} and extracts the names.
+     * Delegates to {@link #getTypedAttributeNames()} and extracts the names.
      */
     @Override
     public Collection<Attribute<?>> getAttributes() {
@@ -223,7 +221,7 @@ public abstract class AbstractEntity implements Entity {
     private class ValueCollection extends AbstractCollection<Attribute<?>> {
         @Override
         public Iterator<Attribute<?>> iterator() {
-            return Iterators.transform(getAttributeTypedNames().iterator(),
+            return Iterators.transform(getTypedAttributeNames().iterator(),
                                        new Function<TypedName<?>, Attribute<?>>() {
                                            @Nullable
                                            @Override
@@ -236,7 +234,7 @@ public abstract class AbstractEntity implements Entity {
 
         @Override
         public int size() {
-            return getAttributeTypedNames().size();
+            return getTypedAttributeNames().size();
         }
     }
 
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
@@ -35,6 +35,24 @@ import java.util.*;
  * Base class to make it easier to implement entities.
  */
 public abstract class AbstractEntity implements Entity {
+    protected final EntityType type;
+    protected final long id;
+
+    protected AbstractEntity(EntityType t, long eid) {
+        type = t;
+        id = eid;
+    }
+
+    @Override
+    public long getId() {
+        return id;
+    }
+
+    @Override
+    public EntityType getType() {
+        return type;
+    }
+
     @Override
     public Collection<AttributeValue<?>> getAttributeValues() {
         return new ValueCollection();
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
@@ -66,14 +66,14 @@ public abstract class AbstractEntity implements Entity {
     /**
      * {@inheritDoc}
      *
-     * This implementation delegates to {@link #maybeGet(Attribute)}.
+     * This implementation delegates to {@link #maybeGet(TypedName)}.
      */
     @Nonnull
     @Override
-    public <T> T get(Attribute<T> attribute) {
-        T val = maybeGet(attribute);
+    public <T> T get(TypedName<T> name) {
+        T val = maybeGet(name);
         if (val == null) {
-            throw new NoSuchAttributeException(attribute.toString());
+            throw new NoSuchAttributeException(name.toString());
         } else {
             return val;
         }
@@ -98,21 +98,21 @@ public abstract class AbstractEntity implements Entity {
     /**
      * {@inheritDoc}
      *
-     * This implementation delegates to {@link #get(Attribute)}.
+     * This implementation delegates to {@link #get(TypedName)}.
      */
     @Override
-    public long getLong(Attribute<Long> attribute) {
-        return get(attribute);
+    public long getLong(TypedName<Long> name) {
+        return get(name);
     }
 
     /**
      * {@inheritDoc}
      *
-     * This implementation delegates to {@link #get(Attribute)}.
+     * This implementation delegates to {@link #get(TypedName)}.
      */
     @Override
-    public double getDouble(Attribute<Double> attr) {
-        return get(attr);
+    public double getDouble(TypedName<Double> name) {
+        return get(name);
     }
 
     @Override
@@ -182,12 +182,12 @@ public abstract class AbstractEntity implements Entity {
         @Override
         public Iterator<AttributeValue<?>> iterator() {
             return Iterators.transform(getAttributes().iterator(),
-                                       new Function<Attribute<?>, AttributeValue<?>>() {
+                                       new Function<TypedName<?>, AttributeValue<?>>() {
                                            @Nullable
                                            @Override
-                                           public AttributeValue<?> apply(@Nullable Attribute<?> input) {
+                                           public AttributeValue<?> apply(@Nullable TypedName<?> input) {
                                                assert input != null;
-                                               return AttributeValue.create((Attribute) input, get(input));
+                                               return AttributeValue.create((TypedName) input, get(input));
                                            }
                                        });
         }
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
@@ -54,7 +54,7 @@ public abstract class AbstractEntity implements Entity {
     }
 
     @Override
-    public Collection<AttributeValue<?>> getAttributeValues() {
+    public Collection<Attribute<?>> getAttributeValues() {
         return new ValueCollection();
     }
 
@@ -141,8 +141,8 @@ public abstract class AbstractEntity implements Entity {
         ToStringBuilder tsb = new ToStringBuilder(this, ToStringStyle.SHORT_PREFIX_STYLE);
         tsb.append("type", getType())
            .append("id", getId());
-        for (AttributeValue<?> av: getAttributeValues()) {
-            tsb.append(av.getAttribute().toString(), av.getValue());
+        for (Attribute<?> av: getAttributeValues()) {
+            tsb.append(av.getTypedName().toString(), av.getValue());
         }
         return tsb.toString();
     }
@@ -178,16 +178,16 @@ public abstract class AbstractEntity implements Entity {
         }
     }
 
-    private class ValueCollection extends AbstractCollection<AttributeValue<?>> {
+    private class ValueCollection extends AbstractCollection<Attribute<?>> {
         @Override
-        public Iterator<AttributeValue<?>> iterator() {
+        public Iterator<Attribute<?>> iterator() {
             return Iterators.transform(getAttributes().iterator(),
-                                       new Function<TypedName<?>, AttributeValue<?>>() {
+                                       new Function<TypedName<?>, Attribute<?>>() {
                                            @Nullable
                                            @Override
-                                           public AttributeValue<?> apply(@Nullable TypedName<?> input) {
+                                           public Attribute<?> apply(@Nullable TypedName<?> input) {
                                                assert input != null;
-                                               return AttributeValue.create((TypedName) input, get(input));
+                                               return Attribute.create((TypedName) input, get(input));
                                            }
                                        });
         }
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
@@ -21,6 +21,7 @@
 package org.lenskit.data.entities;
 
 import com.google.common.base.Function;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterators;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
@@ -29,6 +30,8 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import javax.annotation.OverridingMethodsMustInvokeSuper;
+import javax.print.attribute.standard.MediaSize;
 import java.util.*;
 
 /**
@@ -53,8 +56,28 @@ public abstract class AbstractEntity implements Entity {
         return type;
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * Delegates to {@link #getAttributeTypedNames()} and extracts the names.
+     */
     @Override
-    public Collection<Attribute<?>> getAttributeValues() {
+    public Set<String> getAttributeNames() {
+        // TODO Make this more efficient
+        ImmutableSet.Builder<String> bld = ImmutableSet.builder();
+        for (TypedName<?> name: getAttributeTypedNames()) {
+            bld.add(name.getName());
+        }
+        return bld.build();
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * Delegates to {@link #getAttributeTypedNames()} and extracts the names.
+     */
+    @Override
+    public Collection<Attribute<?>> getAttributes() {
         return new ValueCollection();
     }
 
@@ -115,6 +138,26 @@ public abstract class AbstractEntity implements Entity {
         return get(name);
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * This implementation delegates to {@link #get(TypedName)}.
+     */
+    @Override
+    public int getInteger(TypedName<Integer> name) {
+        return 0;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * This implementation delegates to {@link #get(TypedName)}.
+     */
+    @Override
+    public boolean getBoolean(TypedName<Boolean> name) {
+        return false;
+    }
+
     @Override
     public boolean equals(Object obj) {
         if (obj == this) {
@@ -139,9 +182,8 @@ public abstract class AbstractEntity implements Entity {
     @Override
     public String toString() {
         ToStringBuilder tsb = new ToStringBuilder(this, ToStringStyle.SHORT_PREFIX_STYLE);
-        tsb.append("type", getType())
-           .append("id", getId());
-        for (Attribute<?> av: getAttributeValues()) {
+        tsb.append("type", getType());
+        for (Attribute<?> av: getAttributes()) {
             tsb.append(av.getTypedName().toString(), av.getValue());
         }
         return tsb.toString();
@@ -181,7 +223,7 @@ public abstract class AbstractEntity implements Entity {
     private class ValueCollection extends AbstractCollection<Attribute<?>> {
         @Override
         public Iterator<Attribute<?>> iterator() {
-            return Iterators.transform(getAttributes().iterator(),
+            return Iterators.transform(getAttributeTypedNames().iterator(),
                                        new Function<TypedName<?>, Attribute<?>>() {
                                            @Nullable
                                            @Override
@@ -194,7 +236,7 @@ public abstract class AbstractEntity implements Entity {
 
         @Override
         public int size() {
-            return getAttributes().size();
+            return getAttributeTypedNames().size();
         }
     }
 
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
@@ -20,16 +20,26 @@
  */
 package org.lenskit.data.entities;
 
+import com.google.common.base.Function;
+import com.google.common.collect.Iterators;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.util.*;
 
 /**
  * Base class to make it easier to implement entities.
  */
 public abstract class AbstractEntity implements Entity {
+    @Override
+    public Collection<AttributeValue<?>> getAttributeValues() {
+        return new ValueCollection();
+    }
+
     @Override
     public Map<String, Object> asMap() {
         return new MapView();
@@ -108,6 +118,17 @@ public abstract class AbstractEntity implements Entity {
         return hcb.toHashCode();
     }
 
+    @Override
+    public String toString() {
+        ToStringBuilder tsb = new ToStringBuilder(this, ToStringStyle.SHORT_PREFIX_STYLE);
+        tsb.append("type", getType())
+           .append("id", getId());
+        for (AttributeValue<?> av: getAttributeValues()) {
+            tsb.append(av.getAttribute().toString(), av.getValue());
+        }
+        return tsb.toString();
+    }
+
     private class MapView extends AbstractMap<String,Object> {
         @Override
         public Set<Entry<String, Object>> entrySet() {
@@ -136,6 +157,26 @@ public abstract class AbstractEntity implements Entity {
         @Override
         public Set<String> keySet() {
             return getAttributeNames();
+        }
+    }
+
+    private class ValueCollection extends AbstractCollection<AttributeValue<?>> {
+        @Override
+        public Iterator<AttributeValue<?>> iterator() {
+            return Iterators.transform(getAttributes().iterator(),
+                                       new Function<Attribute<?>, AttributeValue<?>>() {
+                                           @Nullable
+                                           @Override
+                                           public AttributeValue<?> apply(@Nullable Attribute<?> input) {
+                                               assert input != null;
+                                               return AttributeValue.create((Attribute) input, get(input));
+                                           }
+                                       });
+        }
+
+        @Override
+        public int size() {
+            return getAttributes().size();
         }
     }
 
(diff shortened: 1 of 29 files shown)


kluver:
Before Code Review:
 1. Checks are failing and the code is out-of-date with master, so directly merging this is going to have to be your problem. 
 2. This entity class sounds depressingly generic. (I.E. Map or Object, "reimplementing the concept of a class" generic. Without having read the code its not clear to me what we get by having this super-class/interface.

lenskit-core/src/main/java/org/lenskit/data/entities/AttributeValue.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Create a new attribute-value pair.
+     * @param attr The attribute.
+     * @param val The value.
+     * @param <T> The attribute type.
+     * @return The attribute-value object.
+     */
+    public static <T> AttributeValue<T> create(@Nonnull Attribute<T> attr,
```
kluver: Why do you have this if you also have a public constructor?

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.commons.lang3.ClassUtils;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.grouplens.grapht.util.ClassProxy;
+
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.Immutable;
+import java.io.*;
+
+/**
+ * Identifier for an attribute in an entity.
```
kluver: 1. It seems strange for the class named `Attribute` to not actually represent an attribute, but rather the identifier of an attribute.
 2. The javadoc seems a little sparse. It says "what" the class is but it doesn't really explain the class. I don't think what I have here is right, but a proposal:

`This class represents the identifier for an attribute in an entity, and is used in accessing values stored in entities. Each entity is essentially a pair of a string name and a type. By coupling the name to its type we can ensure type safe access to the mixed type entity class`

Also, document the type parameter.

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
```diff
@@ -0,0 +1,157 @@
...
+    private final Class<T> type;
+    private transient volatile int hashCode;
+
+    private Attribute(String n, Class<T> t) {
+        name = n;
+        type = t;
+    }
+
+    /**
+     * Get the field's name.
```
kluver: The use of `field` here is confusing since we don't talk about fields in the class level javadoc. This again indicates to me that there might be a naming issue with this class.

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
```diff
@@ -0,0 +1,157 @@
...
+     * @return The field's name.
+     */
+    @Nonnull
+    public String getName() {
+        return name;
+    }
+
+    /**
+     * Get the field's type.  Primitive field types are normalized to their wrapper types, so this is always a reference
```
kluver: It might just be my personal java terminology literacy at play here but it took me a google search to correctly parse `reference type`. Maybe its that the word `reference` is a bit overloaded in my mind. At any rate, I think it would be more direct to say `this is never a primitive type`

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
```diff
@@ -0,0 +1,157 @@
...
+        return type;
+    }
+
+    @Override
+    public int hashCode() {
+        // intermediate variable means only 1 memory fence in common case
+        int hc = hashCode;
+        if (hc == 0) {
+            HashCodeBuilder hcb = new HashCodeBuilder();
+            hashCode = hc = hcb.append(name)
```
kluver: This feels like a code style issue to me. I'm not sure it is, but it feels like one to me. Beyond looking weird it also made it hard for me to find variable assignments since I don't think coders expect multiple assignments per line.

lenskit-core/src/main/java/org/lenskit/data/entities/AttributeValue.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+
+        if (o == null || getClass() != o.getClass()) return false;
+
+        AttributeValue<?> that = (AttributeValue<?>) o;
+
+        return new EqualsBuilder()
```
kluver: Is it worth standardizing on equalsBuilder? (or standardizing on not equalsBuilder?)

lenskit-core/src/main/java/org/lenskit/data/entities/BuiltBy.java
```diff
@@ -0,0 +1,39 @@
...
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.data.entities;
+
+import org.lenskit.data.events.EventBuilder;
```
kluver: I don't normally comment on imports. Where is this import used?

lenskit-core/src/main/java/org/lenskit/data/entities/BuiltBy.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.lang.annotation.*;
+
+/**
+ * Identify the builder type for an entity class.
+ */
+@Documented
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface BuiltBy {
```
kluver: Do we need this? I feel like we already have annotations that are basically this.

lenskit-core/src/main/java/org/lenskit/data/entities/CommonAttributes.java
```diff
@@ -0,0 +1,61 @@
...
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.data.entities;
+
+/**
+ * Definitions and utilities for common fields.
+ */
+public final class CommonAttributes {
```
kluver: I'm worried that we are moving towards a pattern where this class needs to be updated every time we make a new event type. ID, user, and item are reasonable because they very crosscutting in our domain. Timestamp is a little sketchier, but I could see it being used a lot. Name, Rating, and Count seem more specific to only one entity type. I'm not sure I want those in a central location.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Set;
+
+/**
+ * Base class for data entities in LensKit.  The LensKit data model consists of various entities, each of which has a
+ * type, a numeric identifier (as a `long`), and additional data in a key-value mapping.
+ *
+ * Two entities are equal if their types, IDs, and data are equal.
+ *
+ * Identifiers are scoped per-type; it is acceptable to LensKit for two entities with different types to have the same
+ * identifier, and they are not considered to be the same entity.  However, individual data storage facilities are free
```
kluver: That "However" line isn't clear to me.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Base class for data entities in LensKit.  The LensKit data model consists of various entities, each of which has a
+ * type, a numeric identifier (as a `long`), and additional data in a key-value mapping.
+ *
+ * Two entities are equal if their types, IDs, and data are equal.
```
kluver: (I expect Ill find an answer as I continue reading) Why can't Type and ID be used as an ID on their own?

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
```diff
@@ -0,0 +1,154 @@
...
+     * Get the type of this entity.
+     *
+     * @return The entity's type.
+     */
+    EntityType getType();
+
+    /**
+     * Get the names of the fields in this entity.
+     */
+    Set<String> getAttributeNames();
```
kluver: (This may again be a "answer as I continue review" point) At no point does this seem to be stated, but it seems that within an entity attribute names are keys (not name and type which is the identity of the attribute type (a decision I agree with)) I feel like this distinction should be documented somewhere, `Attribute` is not the `key` to the attribute map, names are.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Get the fields in this entity.
+     */
+    Set<Attribute<?>> getAttributes();
```
kluver: All these different views (as a set, collection, or map) are neat. How hard are they going to be to maintain and are we actually going to need them?

lenskit-core/src/main/java/org/lenskit/data/entities/AttributeValue.java
mdekstrand: Good question... habit... I feel like we haven't fully standardized static factory methods vs. public constructors for these cases. I like the factory methods myself.

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
```diff
@@ -0,0 +1,157 @@
...
+ * Identifier for an attribute in an entity.
+ */
+@Immutable
+public final class Attribute<T> implements Serializable {
+    private static final long serialVersionUID = -1L;
+    private static final Interner<Attribute<?>> FIELD_CACHE = Interners.newStrongInterner();
+
+    private final String name;
+    private final Class<T> type;
+    private transient volatile int hashCode;
```
mdekstrand: It might not be necessary, but it gives us a robust memory model with respect to reads and writes. In practice, because it's 32-bit, we can probably get away without `volatile`, but I *think* that it is, strictly speaking, necessary for the use of this field as a thread-safe cache to be well-defined.

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
mdekstrand: Yep, it should be 'attribute' or some such.

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
mdekstrand: In general, I don't have a problem with using 'reference type', since it is the correct Java term for the thing in question. However, I agree that this case can be rewritten to be easier to read without it, since the relevant point is that it will always return the wrapper class.

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
mdekstrand: I don't like doing it very much. I can change this line to not do it.

lenskit-core/src/main/java/org/lenskit/data/entities/AttributeValue.java
mdekstrand: I tend to use `EqualsBuilder` for all new code, unless the class is trivial. This class may or may not be trivial.

Are you seeing this as something that needs a change here, or something that we should document in the code guidelines?

lenskit-core/src/main/java/org/lenskit/data/entities/BuiltBy.java
mdekstrand: It isn't. I'll remove it.

lenskit-core/src/main/java/org/lenskit/data/entities/BuiltBy.java
mdekstrand: Yes, for code that isn't in this PR. The entity builders are used by file loading code and by the view type transmogrification logic.

You'll be able to tell the DAO 'give me the `rating` entities, and I want to see them as type `Rating`', and if the objects aren't already `Rating` objects, it will look at `Rating`'s `@BuiltBy` to figure out the builder class to use, and use that to convert the object to a `Rating`.

lenskit-core/src/main/java/org/lenskit/data/entities/CommonAttributes.java
mdekstrand: My basic intuition behind what to put here is 'standard attributes for entity types that are in `CommonTypes`', and/or 'the attributes that code included in LensKit itself tries to use'.

But we could move `Rating` off to the `ratings` package if that would make more sense. Not sure where to put it there.

mdekstrand: If I make this move, I won't do it until I get to the point in this process where I am turning `Rating` into an entity.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
```diff
@@ -0,0 +1,154 @@
...
+     *         incompatible type.
+     */
+    @Nonnull
+    <T> T get(Attribute<T> attribute);
+
+    /**
+     * Get the value of an attribute by name.
+     * @param attr The attribute name.
+     * @throws NoSuchAttributeException if the specified attribute is not present.
+     * @return The field value, or `null` if the attribute is missing.
```
kluver: 110 and 111 seem to be contradicting each other.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
mdekstrand: Basically, you can implement a DAO where all entities have to have different IDs, regardless of type, if you feel like it, and you won't be breaking `Entity`'s contract.

Do I even need to mention that, or is it sufficiently clear that the fact that it's acceptable to LensKit does not imply it's acceptable to every application that might implement a LensKit DAO?

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
```diff
@@ -0,0 +1,154 @@
...
+     */
+    boolean hasAttribute(String name);
+
+    /**
+     * Check if the entity has a attribute.
+     *
+     * @param attribute The attribute to look for.
+     * @return `true` if the entity contains `attribute` (same name **and type**).
+     */
+    boolean hasAttribute(Attribute<?> attribute);
```
kluver: Is this interesting? All entities have a statically defined set of attributes right?

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
kluver: and 111 contradicts 113

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
```diff
@@ -0,0 +1,154 @@
...
+     *         incompatible type.
+     */
+    @Nullable
+    <T> T maybeGet(Attribute<T> attribute);
+
+    /**
+     * Get the value of a possibly-missing attribute by name.
+     * @param attr The attribute name.
+     * @return The attribute's value, or `null` if it is not present.
+     * @throws IllegalArgumentException if an attribute with the same name as `attr` is present, but it is of an
```
kluver: not relevant here

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
mdekstrand: good question...

If you have two Entity objects with the same type and ID but different contents, you will probably have a sad time. The key question is what *equality* for entities should mean. I can make an argument either way. LensKit code will generally not depend on the meaning, but I want to make sure that it is defined.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
```diff
@@ -0,0 +1,154 @@
...
+    Object maybeGet(String attr);
+
+    /**
+     * Get the value of a attribute that contains a long.
+     * @param attribute The attribute.
+     * @return The attribute's value.
+     * @throws NoSuchAttributeException if the specified attribute is not present.
+     * @throws IllegalArgumentException if the attribute is present but its type is not `Long`.
+     */
+    long getLong(Attribute<Long> attribute);
```
kluver: why only long and double, why not int and/or boolean. (I have no interest in floats)

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Get the value of a attr that contains a double.
+     * @param attr The attr.
+     * @return The attr's value.
+     * @throws NoSuchAttributeException if the specified attr is not present.
+     * @throws IllegalArgumentException if the attr is present but is not of type `Double`.
```
kluver: not relevant as written. attribute's identify on type so non-Double would mean that its not present

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
kluver: What if the specified attribute is present but null?

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
mdekstrand: Oof.

There are two ways to think about it. One is that way. I think I'm probably OK with it.

The way I was thinking of it is that the (name,type) pair is the ID, with the restriction that no entity can have two attributes with the same name. That's probably weird.

If we go with 'names are keys', we need to be careful how we document the relationship between names and types.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
mdekstrand: `AbstractEntity` will take care of the bulk of the busywork for that. New entity implementations will implement `getAttributes()` and the `get`/`set` methods, basically.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Get the value of a possibly-missing attribute.
+     * @param attribute The attribute.
+     * @param <T> The attribute's type.
+     * @return The attribute's value, or `null` if it is not present.
+     * @throws IllegalArgumentException if a attribute with the same name as `attribute` is present, but it is of an
+     *         incompatible type.
+     */
+    @Nullable
+    <T> T maybeGet(Attribute<T> attribute);
```
kluver: Generics with classes confuse me. therefore I cannot predict what would happen if 
 1. You had Attribute<Class> and Attribute<SubClass> or Attribute<SuperClass> in these contexts.
 2. You have Attribute<Class> and an instance of SubClass

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
mdekstrand: Yes, they do contradict each other.

lenskit-core/src/main/java/org/lenskit/data/entities/EntityBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.data.entities;
+
+/**
+ * Base class for entity builders.  The standard implementation is {@link BasicEntityBuilder}, which can be created
```
kluver: ... what exactly is an entity builder. is it a provider<Entity>? Is there nuance not captured in the name.

lenskit-core/src/main/java/org/lenskit/data/entities/EntityBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+     * Construct a new entity builder.
+     *
+     * **Note:** This constructor calls {@link #setId(long)} and {@link #setType(EntityType)}, so if those methods
+     * are overloaded, they should be able to run without the derived class constructor having finished.
+     *
+     * @param initId The initial ID.
+     * @param initIdSet Whether the ID is initially set.
+     * @param typ The entity type.
+     */
+    protected EntityBuilder(long initId, boolean initIdSet, EntityType typ) {
```
kluver: Its not clear to me why you would want an initial id or entity type, especially since entity type is mutable (its not clear to me why entity type is mutable)

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
mdekstrand: They don't - once we have this all in place, you can set up MovieLens, so `item` entities have `id`, `imdbId`, `title` (or `name`); or BookCrossing, so `item`s have `isbn`, `title`, etc. Or Qian can set up whatever attributes he wants to be able to feed in to his SVDFeature recommender.

One of the goals of entities is to make LensKit data objects extensible without having to write a bunch of new classes all the time.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
mdekstrand: So for this, and another comment or two, we have to think carefully about we want name, type, and 'key' to work together.

lenskit-core/src/main/java/org/lenskit/data/entities/EntityType.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Get the entity type for a name.
+     * @param name The type name.
+     * @return The entity type with name `name`.
+     */
+    @Nonnull
+    @JsonCreator
+    public static synchronized EntityType forName(String name) {
+        String normedName = name.toLowerCase(Locale.ROOT).intern();
+        EntityType type = TYPE_CACHE.get(normedName);
```
kluver: why is this implementing pseduo-singleton (thats not the right label but oyu know what I mean) differently than attribute?

lenskit-core/src/main/java/org/lenskit/data/entities/EntityType.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * A type for an entity.  Obtain an entity from a named type with {@link #forName(String)}.
+ *
+ * Type names are *case-insensitive*, and are normalized to lowercase in the {@linkplain Locale#ROOT root locale}.
+ */
+public final class EntityType implements Serializable {
```
kluver: So far I've read an abstract definition of "object" defined in java. now I feel like I'm reading an abstract definition of "class" defined in java. If it pays off thats fine, I'm just saying things are getting meta

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
mdekstrand: Attributes cannot be present and `null` - they are either present or absent. Less error-prone that way IMO.

`int` is a good idea. I can also be persuaded about `boolean`. And yes, we don't need `float`s, and since we aren't floating, we don't need `short`s, and the fish won't `byte`.

lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
```diff
@@ -0,0 +1,218 @@
...
+           .append(asMap());
+        return hcb.toHashCode();
+    }
+
+    @Override
+    public String toString() {
+        ToStringBuilder tsb = new ToStringBuilder(this, ToStringStyle.SHORT_PREFIX_STYLE);
+        tsb.append("type", getType())
+           .append("id", getId());
+        for (AttributeValue<?> av: getAttributeValues()) {
```
kluver: given that I remember seeing a definition of identifier as an attribute, how likely is is that ID shows up twice?

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
mdekstrand: Oof. Yeah. *Generally*, you won't be dealing with subclasses in attributes. But you might. The goal is that you can safely use this method if `attribute.getType().cast(realValue)` works.

I think we can design around this by adopting your intuition that only the attribute name is considered as the key into the map, and the type associated with the attribute is used as a convenience for accessing/setting. Then we have can have the following rules for accessing an attribute of type `X`:

- If `T.cast(value)` works, it is returned.
- If `T` is `String`, then the attribute is converted with `toString()`
- If the value is a string, then it is converted with Joda-Convert
- If the value and `T` are both numbers, *and* Java would silently convert the value to type `T` in code (e.g. `int`->`long` or `int`->`double`), then it works.

I'm not convinced of the usefulness of the complexity of this last rule. But it is something to consider.

We would implement this logic in `Attribute`. It already knows how to convert strings, we can teach it how to do the rest of the coercions.

lenskit-core/src/main/java/org/lenskit/data/entities/BareEntity.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Implement a bare entity that has no attributes.
+ */
+@Immutable
+class BareEntity extends AbstractEntity {
+    private final EntityType type;
+    private final long id;
+
+    public BareEntity(EntityType t, long eid) {
```
kluver: Doesn't every entity have a type and id? how is this not part of the abstract class?

lenskit-core/src/main/java/org/lenskit/data/entities/EntityBuilder.java
mdekstrand: It is probably needless complexity.

I don't know that we have a use case for `type` being mutable. I can make it immutable and we can change that later if we have a reason.

lenskit-core/src/main/java/org/lenskit/data/entities/BareEntity.java
```diff
@@ -0,0 +1,84 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * Implement a bare entity that has no attributes.
+ */
+@Immutable
+class BareEntity extends AbstractEntity {
```
kluver: As someone who is suspicious of a three layer type hierarchy (interface => abstract => impl) This four layer hierarchy bugs me (interface => abstract => base class => impl)

lenskit-core/src/main/java/org/lenskit/data/entities/EntityBuilder.java
mdekstrand: I'm trying to standardize away from calling providers builders. Still need to make this reality in the LensKit code, but going forward, builders are *not* providers.

Builders are used for verbose, incremental construction of (generally immutable) objects. In some cases, such as `EntityBuilder`, they can be used to give other code the ability to build arbitrary types of objects (for example, `RatingBuilder` will implement `EntityBuilder`, so the same CSV-parsing code can build `Rating` objects and `BasicEntity` and `LikeCount` and whatever).

lenskit-core/src/main/java/org/lenskit/data/entities/BareEntity.java
kluver: Or maybe I misunderstand what this class is fore since I see a basic entity on the horizen

lenskit-core/src/main/java/org/lenskit/data/entities/EntityType.java
mdekstrand: good question.

premature optimization to avoid allocating excess entity type objects required by the interner.

lenskit-core/src/main/java/org/lenskit/data/entities/EntityType.java
mdekstrand: just wait until you get to EntityMetaObjectProtocol.

mdekstrand: or not. I'm not really that sadistic.

lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
mdekstrand: it will show up twice. this code is broken.

mdekstrand: The `ENTITY_ID` attribute was added on later.

lenskit-core/src/main/java/org/lenskit/data/entities/BareEntity.java
mdekstrand: Yes, it does, and moving it up is probably not a bad idea.

lenskit-core/src/main/java/org/lenskit/data/entities/BasicEntity.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Set;
+
+/**
+ * Implementation of an entity backed by a basic set of attributes.
+ */
+@Immutable
+class BasicEntity extends AbstractEntity {
+    private final EntityType type;
+    private final long id;
+    private final Map<Attribute<?>, Object> attributes;
```
kluver: I think things might be more efficient (at least string things) if you store this as Map<String, AtrributeValue<?>>

lenskit-core/src/main/java/org/lenskit/data/entities/BasicEntity.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public BasicEntity(EntityType t, long eid, Map<Attribute<?>, Object> attrs) {
```
kluver: ... Is this actually fully correct? This interface could construct an entity that doesn't strictly conform with the interface (I think) since you could pass multiple attributes with different types in, but the interface assumes that two attributes with one name but two types is forbidden.

lenskit-core/src/main/java/org/lenskit/data/entities/BasicEntity.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Nullable
+    @Override
+    public <T> T maybeGet(Attribute<T> attribute) {
+        return attribute.getType().cast(attributes.get(attribute));
+    }
+
+    @Nullable
+    @Override
+    public Object maybeGet(String attr) {
```
kluver: In light of the above point (name not unique) there are uses of this class that would lead to nonspecified behavior.

lenskit-core/src/main/java/org/lenskit/data/entities/BasicEntity.java
```diff
@@ -0,0 +1,101 @@
...
+        assert id == (Long) attributes.get(CommonAttributes.ENTITY_ID);
+    }
+
+    @Override
+    public long getId() {
+        return id;
+    }
+
+    @Override
+    public EntityType getType() {
```
kluver: Ohh... I see this could be the backing implementation of _any_ entity. But that only works if you pass around untyped entities and perform type checking yourself.

lenskit-core/src/main/java/org/lenskit/data/entities/BasicEntityBuilder.java
```diff
@@ -0,0 +1,87 @@
...
+        this(type, -1, false, new HashMap<Attribute<?>, Object>());
+    }
+
+    BasicEntityBuilder(EntityType typ, long initId, boolean initIdSet, Map<Attribute<?>, Object> attrs) {
+        super(initId, initIdSet, typ);
+        attributes = attrs;
+    }
+
+    @Override
+    public <T> EntityBuilder setAttribute(Attribute<T> attr, T val) {
```
kluver: no checking of name vs. name+type here.

lenskit-core/src/main/java/org/lenskit/data/entities/BasicEntity.java
kluver: Should this be public? or should it be private/protected forcing all construction to go through a Builder

lenskit-core/src/main/java/org/lenskit/data/entities/Entities.java
```diff
@@ -0,0 +1,76 @@
...
+        return new BasicEntityBuilder();
+    }
+
+    /**
+     * Create a new basic entity builder.
+     * @param type The entity type.
+     * @return The entity builder.
+     */
+    public static EntityBuilder newBuilder(EntityType type) {
+        return new BasicEntityBuilder(type, -1, false, new HashMap<Attribute<?>, Object>());
```
kluver: I feel like there was a constructor for this in a different class. Is this only for indirection?

lenskit-core/src/main/java/org/lenskit/data/entities/BasicEntity.java
mdekstrand: Hmm. `Attribute` has speed advantages - it uses reference equality - but I haven't profiled that, and we're going to need to think about how this relates to rethinking the name/type/attribute relationship anyway.

lenskit-core/src/main/java/org/lenskit/data/entities/BasicEntity.java
mdekstrand: It is already package-private, because the class is package-private, but I can drop `public` to make that clearer.

lenskit-core/src/main/java/org/lenskit/data/entities/BasicEntityBuilder.java
mdekstrand: you're right and that will make the users sad.

lenskit-core/src/main/java/org/lenskit/data/entities/BuiltBy.java
kluver: Did I miss where this is used BTW?

lenskit-core/src/main/java/org/lenskit/data/entities/Entities.java
mdekstrand: Not sure what you mean by 'indirection'. I just like to be able to write

```
Entities.newBuilder(CommonTypes.RATING)
        .setId(3028)
        .set(USER, 30)
        .set(ITEM, 20)
        .set(RATING, 3.5)
        .build();
```

I would prefer to write `Entity.newBuilder`, but Java 7 does not allow static methods on interfaces.

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
kluver: So the concern is a simultaneous read and write of the hash code where the write tears and the read doesn't see 0 so it doesn't fully recompute but it also uses the wrong hash code. groovy.

Is it mathematically possible to generate a hash code of 0?

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
kluver: you don't like changing it or doing what you did?

just set to hc and then do your atomic (volatile) write to hashCode on the next line.

lenskit-core/src/main/java/org/lenskit/data/entities/AttributeValue.java
kluver: I like constructors in general but I understand we can't always have them. I would let your YAGNI be your guide about the types of changes and techniques that _require_ a factory method. I would also pick one and go with it since the constructure undermines the static factory as I understand it.

lenskit-core/src/main/java/org/lenskit/data/entities/AttributeValue.java
kluver: I read two classes in a row each of comperable complexity (2 attributes) one that used EqualsBuilder and one that didn't. I was just confused. I don't think it needs a change or a formal policy, it just stood out to me as unussal.

lenskit-core/src/main/java/org/lenskit/data/entities/BasicEntity.java
kluver: didn't notice that, I generally don't expect the Spanish inquisition... I mean nonpublic class. Kinda a moot point.

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
mdekstrand: Yep, that's the issue. Isn't multithreaded programming fantastic?

No, hash codes can never be 0. At least not with our hashing algorithm.

lenskit-core/src/main/java/org/lenskit/data/entities/CommonAttributes.java
kluver: Thats fair, and I was immaging entities having their own class files for this kind of definition in the long run when I commented. Either way though we need to establish a patter for entity names in multiple places since we can't in general assume that people building new code will want to / be able to put things in this file.

lenskit-core/src/main/java/org/lenskit/data/entities/Attribute.java
mdekstrand: I don't like doing what I did. This is about the only kind of place where I do. I have changed in my touchups that I'll push in a bit.

lenskit-core/src/main/java/org/lenskit/data/entities/BuiltBy.java
mdekstrand: No - I haven't pushed the code that uses it.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
kluver: yah you don't need to mention that.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
kluver: Documentation and enforcement are both lacking + important here.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
kluver: It was a lot of code to manage these (and tis the kind of code I struggle to review for bugs) but as long as we think its right and don't have to update it much or ever its no big deal.

lenskit-core/src/main/java/org/lenskit/data/entities/Entity.java
kluver: if you disallow null valued attributes you may make it difficult to implement some things while also maintaining promises about which attributes are present. I would definitly do int since you have an int attribute already. Also if you implement that you must enforce it somewhere (is that maybe in attributeValue?)

lenskit-core/src/main/java/org/lenskit/data/entities/EntityType.java
mdekstrand: Also: `Interner` requires working `hashCode` and `equals` methods. Using my own interning logic lets me not implement those, in favor of using object identity.

lenskit-core/src/main/java/org/lenskit/data/entities/EntityBuilder.java
kluver: Cool.

lenskit-core/src/main/java/org/lenskit/data/entities/EntityType.java
kluver: If I'm going to learn a pattern for non-standard type semantics I only want to learn one pattern for non-standard type semantics.

mdekstrand: I'm changing it to use `Interner`. This optimization was ridiculously premature.

lenskit-core/src/main/java/org/lenskit/data/entities/Entities.java
kluver: The indirection of not using BasicEntityBuilder directly

lenskit-core/src/main/java/org/lenskit/data/entities/CommonAttributes.java
mdekstrand: I agree. Will think about this more when I actually implement the `Rating` logicl

mdekstrand:
With 6c94a90, I am planning to merge this PR.

PR #919, opened 2016-06-21
Memory allocation performance improvements

mdekstrand:
This set of changes improves the performance of recommendation and evaluation, particularly in item-item CF, with the following changes:

- Improve top-*N* scored item accumulator to remove spurious boxing of IDs and scores
- Compute top-N recommendations without creating result objects when no details are required
- Allow top-N metrics to be *list-only*, meaning that they do not require the actual result objects
- Improve the `recommend` eval task to only compute detailed recommendations if at least one metric requires `Result` objects
- Improve the item-item recommender to be able to compute non-detailed recommendations without allocating `Result` objects

### DIFF
--- a/lenskit-core/src/main/java/org/grouplens/lenskit/util/TopNScoredItemAccumulator.java
+++ b/lenskit-core/src/main/java/org/grouplens/lenskit/util/TopNScoredItemAccumulator.java
@@ -24,10 +24,7 @@ import com.google.common.primitives.Doubles;
 import it.unimi.dsi.fastutil.doubles.DoubleArrayList;
 import it.unimi.dsi.fastutil.ints.AbstractIntComparator;
 import it.unimi.dsi.fastutil.ints.IntHeapPriorityQueue;
-import it.unimi.dsi.fastutil.longs.Long2DoubleMap;
-import it.unimi.dsi.fastutil.longs.Long2DoubleMaps;
-import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
-import it.unimi.dsi.fastutil.longs.LongSet;
+import it.unimi.dsi.fastutil.longs.*;
 import org.grouplens.lenskit.collections.CompactableLongArrayList;
 import org.grouplens.lenskit.scored.ScoredId;
 import org.grouplens.lenskit.scored.ScoredIdListBuilder;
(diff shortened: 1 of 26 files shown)


lenskit-core/src/main/java/org/grouplens/lenskit/util/TopNScoredItemAccumulator.java
```diff
@@ -220,13 +217,33 @@ public LongSet finishSet() {
...
-            longs.add(items.get(heap.dequeue()));
+            longs.add(items.getLong(heap.dequeue()));
         }
         clear();
 
         return longs;
     }
 
+    @Override
+    public LongList finishList() {
```
mdekstrand: This is a copy of the already-existing code to finish as a list of scored IDs.

lenskit-eval/src/main/java/org/lenskit/eval/traintest/recommend/RecommendEvalTask.java
```diff
@@ -303,13 +305,17 @@ public ConditionEvaluator createConditionEvaluator(AlgorithmInstance algorithm,
             return null;
         }
 
+        boolean useDetails = tlb != null; // writing requires details
```
kluver: This doesn't make sense to me (comment not helpful). Under what conditions is tlb null or not?

mdekstrand: `tlb` (table layout builder) is null when we are are not writing the recommendations to a file; it is non-null when we are (and therefore require details so we can get a score).

kluver: Definitely bike-shedding.
`//when not null we are writing detailed recommendation file`?

mdekstrand: fine i will paint the bike shed taupe. :stuck_out_tongue:

mdekstrand: fixed.

PR #923, opened 2016-06-28
Add entity view types

mdekstrand:
This adds support for entity view types for #845, making `Rating` an entity.

### DIFF
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
@@ -57,6 +57,16 @@ public abstract class AbstractEntity implements Entity {
     /**
      * {@inheritDoc}
      *
+     * This implementation delegates to {@link #hasAttribute(String)}
+     */
+    @Override
+    public boolean hasAttribute(TypedName<?> name) {
+        return hasAttribute(name.getName());
+    }
+
+    /**
+     * {@inheritDoc}
+     *
      * Delegates to {@link #getTypedAttributeNames()} and extracts the names.
      */
     @Override
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
@@ -61,7 +61,8 @@ public abstract class AbstractEntity implements Entity {
      */
     @Override
     public boolean hasAttribute(TypedName<?> name) {
-        return hasAttribute(name.getName());
+        Object value = maybeGet(name.getName());
+        return value != null && name.getType().isInstance(value);
     }
 
     /**
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
@@ -153,7 +153,7 @@ public abstract class AbstractEntity implements Entity {
      */
     @Override
     public int getInteger(TypedName<Integer> name) {
-        return 0;
+        return get(name);
     }
 
     /**
@@ -163,7 +163,22 @@ public abstract class AbstractEntity implements Entity {
      */
     @Override
     public boolean getBoolean(TypedName<Boolean> name) {
-        return false;
+        return get(name);
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * This implementation delegates to {@link #maybeGet(String)} and checks the type.
+     */
+    @Nullable
+    @Override
+    public <T> T maybeGet(TypedName<T> name) {
+        try {
+            return name.getType().cast(maybeGet(name.getName()));
+        } catch (ClassCastException e) {
+            throw new IllegalArgumentException(e);
+        }
     }
 
     @Override
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/DefaultEntityType.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/DefaultEntityType.java
@@ -1,0 +1,37 @@
+/*
+ * LensKit, an open source recommender systems toolkit.
+ * Copyright 2010-2014 LensKit Contributors.  See CONTRIBUTORS.md.
+ * Work on LensKit has been funded by the National Science Foundation under
+ * grants IIS 05-34939, 08-08692, 08-12148, and 10-17697.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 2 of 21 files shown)


lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
```diff
@@ -57,6 +57,16 @@ public EntityType getType() {
     /**
      * {@inheritDoc}
      *
+     * This implementation delegates to {@link #hasAttribute(String)}
+     */
+    @Override
+    public boolean hasAttribute(TypedName<?> name) {
+        return hasAttribute(name.getName());
```
kluver: Its not clear to me that this is correct. Wouldn't you want to check name and type?

kluver: Hmm this is compatible with API. That said the API has no clean way of verifying that an entity has an attribute of the correct name and type. Most of the time that shouldn't be a problem but I could imagine wanting to duck type entities from time to time.

lenskit-core/src/main/java/org/lenskit/data/entities/DefaultEntityType.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Documented
+public @interface DefaultEntityType {
+    /**
+     * The name of the entity type that is customarily used for this entity view class.
+     * @return The entity type name.
+     */
+    String value();
```
kluver: Why Not EntityType?

lenskit-core/src/main/java/org/lenskit/data/ratings/Rating.java
```diff
@@ -38,16 +40,32 @@
  *
  * @compat Public
  */
+@org.lenskit.data.events.BuiltBy(RatingBuilder.class)
```
kluver: Is there a compelling reason this has to be fully qualified?

kluver: never mind, we have two builtby annotations right now don't we?

lenskit-core/src/main/java/org/lenskit/data/ratings/Rating.java
```diff
@@ -129,7 +153,9 @@ public long getTimestamp() {
      * returns null.
      *  
      * @return {code true} if there is a rating (the preference is non-null).
+     * @deprecated Unrates have gone away.
      */
+    @Deprecated
```
kluver: How long are we keeping unrate code? In my mind we cut them out wholesale for LK3.0

lenskit-core/src/main/java/org/lenskit/data/ratings/Rating.java
```diff
@@ -143,6 +169,51 @@ public double getValue() {
...
+            return timestamp > -1;
+        default:
+            return false;
+        }
+    }
+
+    @Nullable
+    @Override
+    public <T> T maybeGet(TypedName<T> name) {
+        return name.getType().cast(maybeGet(name.getName()));
```
kluver: I think this will produce the wrong exception type if someone asks for (example) timestamp as a float. I think it would produce `ClassCastException` but it should produce `IllegalArgumentException`

lenskit-core/src/main/java/org/lenskit/data/ratings/Rating.java
```diff
@@ -143,6 +169,51 @@ public double getValue() {
         return value;
     }
 
+    @Override
```
kluver: If this is what view objects look like its probably going to make people (who have to implement views) sad. a lot of "mandatory repeat yourself" and strange template code most people won't understand. I don't know how to fix this problem.

kluver: Maybe we could create a abstract viewentity that implements hasAttribute etc. using reflection, maybe cacheing the results of reflection inspection (at a static level) for speedups.

kluver:
A few questions, one "I don't think this is right" (but a pretty pedantic one) and one ask that would have been better caught about two PRs ago.

lenskit-core/src/main/java/org/lenskit/data/entities/DefaultEntityType.java
mdekstrand: Technical limitation of Java annotations.

lenskit-core/src/main/java/org/lenskit/data/ratings/Rating.java
mdekstrand: Yep. One of them will go away.

lenskit-core/src/main/java/org/lenskit/data/ratings/Rating.java
mdekstrand: Yep. That is my plan. I'm trying to introduce entities with minimal initially-forced changes, then update the algorithms, then rip out the unsupported functionality.

lenskit-core/src/main/java/org/lenskit/data/ratings/Rating.java
mdekstrand: You are right. Will fix that.

lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
mdekstrand: Good question. I can go either way on it. It  is not hard to add a type check.

Right now, the API def maintains a relationship that the `get` methods throw `NoSuchAttributeException` if and only if `hasAttribute` returns false. If we make this check type and return `false` on a mismatch, then if `hasAttribute` returns `false` then `get` might throw `NoSuchAttributeException`, or it might throw `IllegalArgumentException`.

mdekstrand: In practice, that shouldn't much matter, because either no-such-attribute or illegal-argument is the result of an entity failing to satisfy the contract required by the code using it.

lenskit-core/src/main/java/org/lenskit/data/ratings/Rating.java
mdekstrand: Good idea.

mdekstrand: I have recorded this idea in #924.

lenskit-core/src/main/java/org/lenskit/data/entities/AbstractEntity.java
kluver: I like the symmetry between `has=false` and `get=NoSuchAttrException` but right now we do not have the symetry that`has=true` then `get` will not fail, which I find more important. As you say, these exceptions should never happen in code and are basically unrecoverable (excepting shenanigans).

mdekstrand: What if we make `NoSuchAttributeException` a subclass of `IllegalArgumentException`?

PR #926, opened 2016-06-29
Various improvements for new data structures

mdekstrand:
This has a grab-bag of improvements for #845 work (new DAO things), and some support code.

This is the *persistent* code that will be used to support #925, so that we can consider sticky changes separately from ephemeral bridge code.


lenskit-core/src/main/java/org/lenskit/util/io/ObjectStreams.java
```diff
@@ -159,18 +161,22 @@ public void close() {
      * @return A new list containing the elements of the object stream.
      */
     @SuppressWarnings("PMD.LooseCoupling")
-    public static <T> ArrayList<T> makeList(@WillClose ObjectStream<? extends T> objectStream) {
-        ArrayList<T> list;
+    public static <T> List<T> makeList(@WillClose ObjectStream<? extends T> objectStream) {
```
mdekstrand: This is the heart of some object stream changes. What we're doing is enabling an object stream that wraps a list to be *unwrapped*, instead of requiring the copy all the time, to fast-path some common operations.

lenskit-core/src/main/java/org/lenskit/data/dao/file/CollectionEntitySource.java
```diff
@@ -26,12 +26,14 @@
...
 
 import java.io.IOException;
+import java.io.Serializable;
 import java.util.Collection;
 
 /**
  * Entity source backed by a collection.
  */
-class CollectionEntitySource implements EntitySource {
+class CollectionEntitySource implements EntitySource, Serializable {
```
qian2015: I think you meant it to be a public class.

qian2015: The use case I'm facing is that I need an interface which gives me a stream of entities. And also, I can restart the stream. This is for iterative training of models. The reason for this design is that algorithms have the control of iteration. I found EntitySource is a good fit. I feel like Source confuses with DAO. Maybe only DAO is enough because it seems that the "file" subfolder in "dao" means a specific file type of daos.

qian2015: I hope this interface is not specific to file because entity data might come from elsewhere.

mdekstrand: The `file` should probably be renamed to `static` - it really encompasses a family static data sources.

For iterative methods, I think you are right that we need something else. Perhaps something like the `PreferenceSnapshot` but more generic. It should live on top of the DAOs, and have a fast path to reuse the in-memory data strategy employed by the static file DAO provider (which a subsequent PR is renaming to `StaticDataSource`).

`EntitySource` is not intended to be a very public component - it's mostly an implementation detail of the static DAO. Algorithms that depend on it will not work with any other DAOs, such as a live database DAO. It is public because the evaluator needs to be able to reach into the internal components of the static DAO provider.

PR #949, opened 2016-09-12
Fix ProgressLogger % issue

kboise:
Hello,

I have created a pull request for the fixed ProgressLogger issue. 

Best,
Kiran


lenskit-core/src/main/java/org/lenskit/util/ProgressLogger.java
```diff
@@ -133,7 +133,7 @@ public void logProgress() {
             double est = (total - ndone) * rate;
             logger.info("{}: finished {} of {} ({}%, {}s/row, ETA {})",
                         label, ndone, total,
-                        String.format("%.2f", ((double) ndone) / total),
+                        String.format("%.2f", (((double) ndone) * 100) / total),
```
mdekstrand: Yep, that is correct!

mdekstrand: We can simplify this, though: multiplying `ndone` by `100.0` (note the decimal point) will convert `ndone` to a double :).

mdekstrand:
Superseded by #955.

PR #953, opened 2016-09-22
Upgrade Gradle to 3.3

mdekstrand:
This PR updates our Gradle build to 3.1.


mdekstrand:
I think we're likely going to sit on this PR a bit - would like to cut `-M2` on Gradle 2 I think.

Issue #954, opened 2016-09-22
Implement bias model and related normalizers


mdekstrand:
As per Google Groups discussion, we want to implement a *bias model* for LensKit, and associated bias normalizers and item scorers.

The bias model will have basic user-item biases, and may support looking up info for additional biases.

This may supersede #846.

mdekstrand:
Finished in #956.

PR #956, opened 2016-10-02
User-item bias model

mdekstrand:
This PR starts adding user-item bias models (#954), to be used as item scorers and normalizers. It is not ready to merge, but I want to better think about whether the design is headed in the right direction.

I do need to add a 'live' user-item model that uses precomputed item means but re-fetches the item means on the fly.

### DIFF
--- a/lenskit-core/src/main/java/org/lenskit/bias/BiasModel.java
+++ b/lenskit-core/src/main/java/org/lenskit/bias/BiasModel.java
@@ -20,6 +20,9 @@
  */
 package org.lenskit.bias;
 
+import it.unimi.dsi.fastutil.longs.Long2DoubleMap;
+import it.unimi.dsi.fastutil.longs.LongSet;
+
 /**
  * Interface for bias models that can be based on the user, item, or both.
  */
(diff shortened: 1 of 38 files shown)


lenskit-core/src/main/java/org/lenskit/bias/BiasModel.java
```diff
@@ -0,0 +1,46 @@
...
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.bias;
+
+/**
+ * Interface for bias models that can be based on the user, item, or both.
+ */
+public interface BiasModel {
```
mdekstrand: This is the main bias model interface.

I'm thinking it'll be useful to have mult-user and multi-item getters as well.

mdekstrand: I have added multi-{user,item} getters 😃

PR #983, opened 2017-02-06
Support compound attribute types

mdekstrand:
These changes will support compound attribute types, such as lists of strings, in the LensKIt data model.


mdekstrand:
I am not fond of the number of `unchecked` warnings I am suppressing in this code. Driving a truck through the type system.

PR #990, opened 2017-02-22
Some cleanups of warnings and deprecations

mdekstrand:


lenskit-cli/src/main/java/org/lenskit/cli/LenskitCommandException.java
```diff
@@ -0,0 +1,46 @@
...
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.cli;
+
+/**
+ * Exception thrown when a LensKit command fails.
+ */
+public class LenskitCommandException extends Exception {
+    public LenskitCommandException() {
```
mdekstrand: This needs documentation.

Issue #986, opened 2017-02-10
Remove SparseVector


mdekstrand:
We want to remove `SparseVector` and friends after #937.

Issue #937, opened 2016-07-08
Update algorithms for LK3 changes


mdekstrand:
We need to update all algorithms to use the new DAOs and to avoid using SparseVector.

- [x] Remove old DAO uses
- [x] Remove old SparseVector uses

mdekstrand:
#984 does the DAO side of this, just need SparseVector.

mdekstrand:
Still probably a few stray things, but this is basically done.

PR #991, opened 2017-02-23
Make user-user CF use new vectors (#986)

mdekstrand:
This works on #986, removing `SparseVector` from user-user CF.


mdekstrand:
Also is working on #937.

PR #997, opened 2017-03-13
Implemented a GreedyRerankingItemRecommender.

kluver:
At this point I would like comments and review of the main algorithm code / documentation / structure.

This PR adds a new item recommender - the GreedyRerankingItemRrcommender which implements the greedy re-scoring / re-ranking approach for recommending commonly used to optimize for set properties of the recommendation list such as diversity.

TODO:
- [x] add tests
- [x] I am unsure about the correct package. My current solution is one option, but other options are possible (in particular, putting this in hybrids either alongside the rank blending recommender, or putting this in hybrids and refactoring hubrid into two subfolders.)
- [x] review names of objects to see if something better occurs
- [x] add helper abstract class for CandidateItemSelector
- [x] A lot of TODO items many of which are questions for the review
- [x] re-do the lines between scoring and greedy item selection.
- [x] license headers
- [x] re-ranking thingy knows user.


lenskit-core/src/main/java/org/lenskit/rerank/Rescorer.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Interface for classes that produce scores for the value of adding an item to a list of recommended items. This class
+ * is used in {@link GreedyRerankingItemRecommender} to define the metric to optimize when recommending items.
+ *
+ * While this interface does not have support for hard filtering, a filter behavior can be approximated by returning
+ * {@link Double#MIN_VALUE}. Filtered items may still be returned, but only after all non-filtered items are eliminated.
+ *
+ * @author Daniel Kluver
+ */
+public interface Rescorer {
```
mdekstrand: I like having this interface. I really don't like its name.

lenskit-core/src/main/java/org/lenskit/rerank/NumCandidates.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.grouplens.grapht.annotation.DefaultDouble;
+import org.lenskit.hybrid.RankBlendingItemRecommender;
+import org.lenskit.inject.Parameter;
+
+import javax.inject.Qualifier;
+import java.lang.annotation.*;
+
+/**
+ * Number of candidate items to get from base recommender. Set to negative for no preference.
```
mdekstrand: Any reason not to just always request all? Or request 2-3 times the specified number, and re-request if needed?

lenskit-core/src/main/java/org/lenskit/rerank/GreedyRerankingItemRecommender.java
```diff
@@ -0,0 +1,77 @@
...
+
+        //TODO: use ResultLists instead of Lists
+        //TODO: is this the best List for the situation?
+
+        List<GreedyRerankingResult> results = new ArrayList<>(n);
+        for (int i = 0; i<n; i++) {
+            // TODO: I would like to make this a private function for better method decomposition, but then I have to recompute scores.
+            Result bestResult = null;
+            double bestScore = 0;
+            for (Result option : candidates) {
```
mdekstrand: Seems like we have an interesting performance issue here.

For the general case, looking at score is good. And we clearly need to scan to find 'best', but it might be helpful to have a 'but you won't find better' signal.

Or maybe the right abstraction boundary is to ask the inner to pick instead of score. That way it can pick more efficiently: for a binary case, it can just use the first one that meets the binary condition.

lenskit-core/src/main/java/org/lenskit/rerank/Rescorer.java
kluver: Sure. Options (feel free to remix until we find truth):
 * `ItemSetScorer`
 * `RecommendationCandidateEvaluator`
 * `GreedyRerankingCriteria`

lenskit-core/src/main/java/org/lenskit/rerank/NumCandidates.java
kluver: 1. Efficiency - This may be pre-mature optimization, but this algorithm is O(n^2) as opposed to (and in addition to) the O(n log (k)) ranking we already do. If we rank over all items it might get expensive so I think we need _some_ system for only taking the best options.
1. Replicate past work - I'm pretty sure past papers have used a constant-n approach. Ziegler used 50 for experiments with top-10 for example. (that said, that could easily be converted to a factor)
2. If we request a factor we should have a parameter for that factor. bringing up back to essentially the same conversation. For lack of a concrete use case I don't want to make it generic between factor and number right now. 
3. Absolute number allows negative numbers covering the "just ask for all" use-case" (so long as the recommender also assumes -1 == all, you left this frustratingly vague in the documentation).

I think the only really general solution is to have that be configureable in a more generic way (to allow "recommend all items above threshold" for example) I don't know if I want that fix to be this PR and I don't know if that part really belongs here or in a more generic level.

lenskit-core/src/main/java/org/lenskit/rerank/GreedyRerankingItemRecommender.java
kluver: Hmm... thats an interesting solution. It allows hard filtering much better and also allows approximate solutions in large item spaces (best over the metric over a O(sqrt(n)) subset of available items).

It certainly is more programing than we need to solve the problem at hand, but it is better. I'll see if I can justify the time to myself. to build that out and the various implementations that come to mind.

lenskit-core/src/main/java/org/lenskit/rerank/NumCandidates.java
kluver: If we make the below change, we could theoretically request all items and make the `NextCandidateItemSelector` take charge of this aspect. A `NextCandidateItemSelector` could easily only select candidates from the top-N unselected items (Trading true optimal for efficiency)

mdekstrand: > Efficiency - This may be pre-mature optimization, but this algorithm is O(n^2) as opposed to (and in addition to) the O(n log (k)) ranking we already do. If we rank over all items it might get expensive so I think we need some system for only taking the best options.

For fully-general score-based ones, yes. However, if we go with the 'pick' approach below, then we may have substantially faster ones. 

> Replicate past work - I'm pretty sure past papers have used a constant-n approach. Ziegler used 50 for experiments with top-10 for example. (that said, that could easily be converted to a factor)

Good argument.

> If we request a factor we should have a parameter for that factor. bringing up back to essentially the same conversation. For lack of a concrete use case I don't want to make it generic between factor and number right now.

Agreed. Either factor or number. Factor makes this more useful in non-experimental settings where the client may provide different list sizes.

> Absolute number allows negative numbers covering the "just ask for all" use-case" 

Can also do that with a factor of 0 or -1.

> (so long as the recommender also assumes -1 == all, you left this frustratingly vague in the documentation).

It does. Actually, any negative value is supposed to be all. However, the docs should be clearer on this point.

kluver: What do you think of my proposal: the main recommender always requests full rank, the new item selector (`NextRecommendedItemSelector`?) is then in charge of any possible efficiency / optimality trade-offs. We loose a little in possibly over-requesting the rank, but its a pretty trivial loss for a lot of added flexibility.

mdekstrand:
Package: we have `lenskit-predict`, what if we have `lenskit-rank`? `lenskit-recommend` feels weird (and maybe `lenskit-predict` is also weird).

kluver:
I was making the change you suggested, but then I backed out. We probably would want to move a lot more than my code (In particular lenskit-core/src/.../hybrid. I think moving that would be a breaking change for MovieLens (Isn't MovieLens on LK3? I know they use that algorithm...).

Thoughts?

mdekstrand:
@kluver I do not know what the best layout is. We should have a `lenskit-hybrid` at some point, that has basic hybrids (and probably move the hybrid from `core` - the package name can stay the same, the Gradle module will just change). So long as packages stay the same, moving Gradle modules is not a breaking change if client code depends on `lenskit-all`.

Do you consider this reranker to be a hybrid or something else?

We could also have `lenskit-postproc`, that collects things that postprocess other recommenders' output (hybrids, ordrec, rerank, etc.). I actually kinda like that.

lenskit-core/src/main/java/org/lenskit/rerank/CandidateItemSelector.java
```diff
@@ -0,0 +1,56 @@
...
+ * Interface for classes that select the next item to recommend from a list of candidate items. This class
+ * is used in {@link GreedyRerankingItemRecommender} to define the greedy strategy for selecting the next item.
+ * Often this will be a search for the best item that satisfies some constraint or the best item by some metric that
+ * takes in account what items have already been recommended.
+ *
+ * @author Daniel Kluver
+ */
+
+
+public interface CandidateItemSelector {
```
mdekstrand: I'm not in love with this name - it overloads with `ItemSelector` (not a big problem), and with the yet-to-be-introduced `CandidateSelector` that will abstract the notion of identifying a default set of candidates for a recommend operation.

What if we call it `GreedyRerankStrategy`, so that the name indicates clearly the use of the Strategy pattern?

lenskit-core/src/main/java/org/lenskit/rerank/GreedyRerankingItemRecommender.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @author Daniel Kluver
+ */
+public class GreedyRerankingItemRecommender extends AbstractItemRecommender {
+    private static final Logger logger = LoggerFactory.getLogger(GreedyRerankingItemRecommender.class);
+    private final ItemRecommender baseRecommender;
+    private final CandidateItemSelector selector;
+
+    @Inject
+    public GreedyRerankingItemRecommender(ItemRecommender baseRecommender, CandidateItemSelector selector) {
```
mdekstrand: And then we could call this `strategy`

lenskit-core/src/main/java/org/lenskit/rerank/GreedyRerankingItemRecommender.java
```diff
@@ -0,0 +1,91 @@
...
+            n = candidates.size();
+        }
+
+        List<Result> results = new ArrayList<>(n);
+        for (int i = 0; i<n; i++) {
+            Result nextItem = selector.nextItem(user, n, results, candidates);
+            if (nextItem == null) {
+                break;
+            } else {
+                //TODO: REVIEW: there should be an easier way to do this.
```
mdekstrand: Yes, there should. I was just looking to see if Guava offers any help, but they just have `removeAll` and `removeIf`, no `removeFirst`.

This might be a hideous abstraction break, but what if the strategy returned an index instead of a result? That wouldn't let it also do the rescoring :(, but it would make it easy to go say 'delete item X'.

lenskit-core/src/main/java/org/lenskit/rerank/RerankingRescoredResult.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * A class to store the original score as well as the objective metric score returned by the
+ * objective function algorithm.
+ *
+ * @see AbstractScoringCandidateItemSelector
+ *
+ * @author Daniel Kluver
+ */
+public class RerankingRescoredResult extends AbstractResult {
```
mdekstrand: Is there a reason we want to specialize this, instead of using the result of [Results.rescore](http://lenskit.org/master/apidocs/org/lenskit/results/Results.html#rescore-org.lenskit.api.Result-double-)?

mdekstrand:
Ok, Guava gives us a kinda ugly way to do the removal. Don't know if it's cleaner or not.

1. Transform the result list into a `LinkedList<Equivalence.Wrapper<Result>>` using an ID eqivalence (`Results.idEquivalence()` would be a good place to define this).
2. Use `list.remove(...)` to remove the result.
3. At each step, transform the list of candidates back in to a list of unwrapped results to pass to the strategy.

Step (3) is still O(n), but if the existing lists being modified are array lists, it's O(n) anyway. But if the list being searched is a linked list, then this does introduce computational overhead. It'd be nice if Java gave us a nice way to 'remove first matching a predicate', but nope.

kluver:
95%? looks like I missed a test :p
(Actually its good that I did check since a mistake on n==0 vs. n<0 would be easy to make)

kluver:
My instinct is to make this PR not include module level restructuring (and since that would be incoming to ignore any relevant package level re-structuring). Module level restructuring of other packages isn't involved in implementation of this one and I don't want to block. (Also module & project structure is  approximately 100% your realm on LensKit)

lenskit-core/src/main/java/org/lenskit/rerank/CandidateItemSelector.java
kluver: I like that name. Will change. Names throughout this project have escaped me, (In particular I've been stumbling over existing names)

lenskit-core/src/main/java/org/lenskit/rerank/GreedyRerankingItemRecommender.java
kluver: 👍

lenskit-core/src/main/java/org/lenskit/rerank/RerankingRescoredResult.java
kluver: Didn't even think to check if that was an option.

lenskit-core/src/main/java/org/lenskit/rerank/GreedyRerankingItemRecommender.java
kluver: I don't like the abstraction break. It would  remove the ability to do _any_ result modifications. While our current use cases are simple I want to leave the door open for complicated side-data if needed.

kluver: RemoveIf will do. Only one recommended item should match (I'm making the (in my opinion) reasonable assumption that the list has no item duplicates). Whenever you go to language level 8 we can transition that into direct removeIf. It scans the whole list which is no less efficient in the worst case but it is less efficient in the common case. If that proves to be a concern/problem we keep/revert to the ugly-but-serviceable DIY code, possibly moving it to a `FooUtil` somewhere.

kluver:
@mdekstrand If you agree with the following:
 * This is not the PR for solving the re-rankers / hybrids / post processors issue (where do they belong, and lets go put them there)
 * Guava's remove_if is good enough, our expected case performance is worse but there does not seem to be a non-ugly fix for that
Then I think we are ready to pull

lenskit-core/src/main/java/org/lenskit/rerank/GreedyRerankingItemRecommender.java
```diff
@@ -70,19 +72,16 @@ protected ResultList recommendWithDetails(long user, int n, @Nullable LongSet ca
...
                 break;
             } else {
-                //TODO: REVIEW: there should be an easier way to do this.
-                // Remove the selected item from the list of candidate items.
-                for(Iterator<Result> it = candidates.iterator(); it.hasNext();) {
-                    Result listResult = it.next();
-                    if (listResult.getId() == nextItem.getId()) {
-                        it.remove();
-                        break;
+                Iterables.removeIf(candidates, new Predicate<Result>() {
```
mdekstrand: We can simplify this by adding an `idEquivalence` to `Results`, and then this becomes:

```java
Iterables.removeIf(candidates, Results.idEquivalence().equalTo(nextItem));
```

However, we can do that as a separate cleanup pass.

PR #998, opened 2017-03-14
Feature/tian slim

KimuraTian:
# slim demo
main classes
AbstractLinearRegression:  Abstract class for elastic linear regression which is implemented by two different coordinate descent updates (CovarianceUpdateCoordDestlinearRegression and NaiveCoordDestLinearRegression)
SlimBuildContext: data context used by slim learning. (item rating map and item-item inner-products map)
SlimModelProvider: using ItemItemModel of knn package to find neighbors, and then training each item ItemItemModel.


mdekstrand:
@KimuraTian Looks like there's a compile error that's causing the build to fail.

KimuraTian:
Thanks for mentioning it. I will check it again.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegression.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.inject.Qualifier;
+import java.lang.annotation.*;
+
+
+@Documented
+@Qualifier
+@DefaultImplementation(CovarianceUpdateCoordDestLinearRegression.class)
+@Target({ElementType.TYPE, ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface LinearRegression {
```
mdekstrand: This interface doesn't seem to be adding anything - qualifiers would get applied to injection points, not to injectable interfaces, so it doesn't seem to be used.

lenskit-slim/src/main/java/org/lenskit/slim/AbstractLinearRegression.java
```diff
@@ -0,0 +1,138 @@
...
+import static org.lenskit.slim.LinearRegressionHelper.addVectors;
+
+
+/**
+ * Linear regression which minimize following loss function
+ * 1/2*||a_j - A*w_j||^2 + beta/2*||wj||^2 + lambda*||w_j||
+ */
+@LinearRegression
+@DefaultImplementation(CovarianceUpdateCoordDestLinearRegression.class)
+public abstract class AbstractLinearRegression {
```
mdekstrand: Since this is - right now - SLIM-specific, what if we called it something SLIM-specific? We'll generalize later.

Maybe call it `SLIMScoringStrategy`?

lenskit-slim/src/main/java/org/lenskit/slim/AbstractLinearRegression.java
```diff
@@ -0,0 +1,138 @@
...
+        return residulsUpdated;
+    }
+
+    /**
+     * compute predictions given test data matrix and learned weights
+     * @param testData column-wise data matrix
+     * @param weights weight
+     * @return prediction vector
+     */
+    public Long2DoubleMap predict(Map<Long, Long2DoubleMap> testData, Long2DoubleMap weights) {
```
mdekstrand: We can have `Long2ObjectMap<Long2DoubleMap>`.

Although really, given how much they're used here and elsewhere, we should probably have a `Long2Long2DoubleMap` or `LLDSparseMatrix` or something like that, that encapsulates the concept. Not that you have to build such a thing for this PR, it just seems like it'd be useful here.

lenskit-slim/src/test/java/org/lenskit/slim/NaiveCoordDestLinearRegrTest.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    @Test
+    public void testData() {
+        int sizeOfData = data.keySet().size();
+        assertThat(sizeOfData, equalTo(10));
+    }
+
+    @Test
+    public void testLinearRegression() {
```
mdekstrand: This test should have some assertions in it.

KimuraTian:
Thank you so much. So far it might be not ready for review, some test cases will be added soon.

lenskit-slim/src/main/java/org/lenskit/slim/CovarianceUpdate.java
```diff
@@ -0,0 +1,196 @@
...
+                        }
+                    }
+                } else {
+                    Long2DoubleMap dotProdsOfXjXks = new Long2DoubleOpenHashMap();
+                    for (long weightsIdk : nonzeroWeights.keySet()) {
+                        double dotProdOfXjXk = Vectors.dotProduct(itemXj, trainingDataMatrix.get(weightsIdk));
+                        dotProdsOfXjXks.put(weightsIdk, dotProdOfXjXk);
+                    }
+                    innerProdsOfXs.put(j, dotProdsOfXjXks);
+                }
```
mdekstrand: can we de-duplicate loop bodies here?

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
```diff
@@ -0,0 +1,143 @@
...
+        return mapT;
+    }
+
+    /**
+     * add two vectors
+     * @param a
+     * @param b
+     * @return
+     */
+    public static Long2DoubleMap addVectors(Long2DoubleMap a, Long2DoubleMap b) {
```
mdekstrand: don't we have this in `Vectors`? If not we should.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
```diff
@@ -0,0 +1,143 @@
...
+
+    }
+
+    /**
+     * element-wise multiplication of two vectors.
+     * @param a
+     * @param b
+     * @return
+     */
+    public static Long2DoubleMap multiply(Long2DoubleMap a, Long2DoubleMap b) {
```
mdekstrand: This should also be in `Vectors`

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Transpose of matrix
+     * Convert a row view of rating matrix to a column view and vice versa
+     * After conversion, the key of empty row/column in the input matrix will be eliminated in the transpose matrix.
+     * So calling transposeMap(transposeMap(X)).keySet() may not equal to X.keySet() unless X has no mapping of key to empty Long2Double map.
+     *
+     * @param matrix
+     * @return
+     */
+    public static Long2ObjectMap<Long2DoubleMap> transposeMap(Long2ObjectMap<Long2DoubleMap> matrix) {
```
mdekstrand: This should perhaps belong in a matrix utility class (to go with `Vectors`).

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
```diff
@@ -0,0 +1,143 @@
...
+        return productOfTwoVectors;
+    }
+
+//    /**
+//     * multiply vector by a real value
+//     * @param a
+//     * @param c
+//     * @return
+//     */
+//    public static Long2DoubleMap multiply(Long2DoubleMap a, double c) {
```
mdekstrand: let's delete this commented-out code.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
```diff
@@ -0,0 +1,143 @@
...
+//        return scalingVector;
+//    }
+
+    /**
+     * return a new map with all values in input map but not equal to the given value
+     * @param a input map
+     * @param value a value needs to be filtered out
+     * @return
+     */
+    public static Long2DoubleMap filterValues(Long2DoubleMap a, double value) {
```
mdekstrand: We might want to introduce a customizable epsilon here? `Scalars.isZero` is pretty restrictive by default.

lenskit-slim/src/main/java/org/lenskit/slim/SLIMScoringStrategy.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * learning process
+     *
+     * @param labels label vector
+     * @param trainingDataMatrix observations matrix row: user ratings for different items, column: item ratings of different users
+     * @return a trained weight vector
+     */
+    public abstract Long2DoubleMap fit(Long2DoubleMap labels, Long2ObjectMap<Long2DoubleMap> trainingDataMatrix);
```
mdekstrand: Do we actually use this one anywhere other than in tests?

lenskit-slim/src/main/java/org/lenskit/slim/UpdateDescentRule.java
```diff
@@ -0,0 +1,39 @@
...
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.slim;
+
+import it.unimi.dsi.fastutil.longs.Long2DoubleMap;
+import java.util.Map;
+
+
+/**
+ * not used. wonder if abstract class is better than interface for different updates??
```
mdekstrand: let's delete it.

mdekstrand:
Also, I don't see clear hooks for normalization right now - should we add those?

lenskit-slim/src/main/java/org/lenskit/slim/CovarianceUpdate.java
KimuraTian: Pulling in an item-item model is just to use its hooks for different similarity strategies. Now the codes for similarity strategies are copied into SLIMBuildContextProvider which will return a mapping from item ids to sets of corresponding neighbors. All the dependencies on KNN are removed.

KimuraTian: Yes, loop bodies are de-duplicated here.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
KimuraTian: This method is not used in SLIM except in test.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
KimuraTian: We have a similar version 'Vectors.combine'. But 'Vectors.combine' may only combine the elements in common positions between two vectors. It may not work correctly for the sparse vector addition.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
KimuraTian: This may also not be used except tests.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
KimuraTian: A customizable epsilon is added

lenskit-slim/src/main/java/org/lenskit/slim/SLIMScoringStrategy.java
KimuraTian: No. It is just kept as an alternative and used as a double check in the tests in case something goes wrong.

lenskit-slim/src/main/java/org/lenskit/slim/CovarianceUpdate.java
```diff
@@ -0,0 +1,190 @@
...
+
+                double weightToUpdate = weights.get(j);
+                double weightUpdated = updateWeight(itemXj, dotProdOfXjY, dotProdsOfXjXks, nonzeroWeights, lambda, beta);
+                weights.put(j, weightUpdated);
+                residuals = updateResiduals(residuals, itemXj, weightToUpdate, weightUpdated);
+                loss[k%2] = computeLossFunction(residuals, weights);
+            }
+            k++;
+            lossDiff = Math.abs(loss[0] - loss[1]); // compute difference of loss function between two round of coordinate descent updates
+            logger.info("{}th iteration and loss function reduced to {} and {} \n and weights is {}", k, loss[0], loss[1], weights);
```
mdekstrand: Let's make this `debug`-level; it's too noisy at `info`.

lenskit-slim/src/main/java/org/lenskit/slim/Epsilon.java
```diff
@@ -0,0 +1,40 @@
...
+ * The tolerance used to throw away zero elements of the learned weight vector
+ * @author <a href="http://www.grouplens.org">GroupLens Research</a>
+ */
+@Documented
+@DefaultDouble(Double.MIN_VALUE)
+@Parameter(Double.class)
+@Qualifier
+@Target({ElementType.METHOD, ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Epsilon {
```
mdekstrand: Let's call this `MinItemWeight`.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
mdekstrand: That does not matter; we can have utilities in those classes that are only used for tests.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
mdekstrand: Good point! We should perhaps have both versions in `Vectors`; maybe `combine` and `combineAndLimit`?

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
mdekstrand: Maybe call it `ebeMultiply` to match the `commons-math` method name? But also, it should live in `Vectors`, even if it is only used in tests.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
```diff
@@ -0,0 +1,130 @@
...
+     */
+    public static Long2DoubleMap multiply(Long2DoubleMap a, Long2DoubleMap b) {
+        Long2DoubleMap productOfTwoVectors = new Long2DoubleOpenHashMap();
+        LongOpenHashBigSet aSet = new LongOpenHashBigSet(a.keySet());
+        LongOpenHashBigSet bSet = new LongOpenHashBigSet(b.keySet());
+        aSet.retainAll(bSet); // intersection of two key sets
+
+        for (long key : aSet) {
+            double prod = a.get(key)*b.get(key);
+            productOfTwoVectors.put(key, prod);
```
mdekstrand: if `prod` is zero, we don't need to put it in the vectors.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
```diff
@@ -0,0 +1,130 @@
...
+
+
+    /**
+     * return a new map with all values in input map but not equal to the given value
+     * @param a input map
+     * @param value a value needs to be filtered out
+     * @param epsilon tolerance
+     * @return a map with the values in {@code a} whose absolute value of differences from {@code value} is greater than {@code epsilon}.
+     */
+    public static Long2DoubleMap filterValues(Long2DoubleMap a, double value, double epsilon) {
```
mdekstrand: This should also live in `Vectors`.

lenskit-slim/src/main/java/org/lenskit/slim/MinCoRatedItems.java
```diff
@@ -0,0 +1,41 @@
...
+ * Items with fewer than this many users in common will be ignored, i.e. treated as having no similarity.
+ * A hard threshold allows for performance optimizations.
+ */
+@Qualifier
+@Documented
+@Parameter(int.class)
+@DefaultInteger(0)
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.PARAMETER, ElementType.FIELD, ElementType.METHOD})
+public @interface MinCoRatedItems {
```
mdekstrand: `knn` has a `MinCommonUsers` parameter - let's reuse that?

lenskit-slim/src/main/java/org/lenskit/slim/Intercept.java
```diff
@@ -0,0 +1,40 @@
...
+ * Whether or not including intercept in linear regression.
+ * @author <a href="http://www.grouplens.org">GroupLens Research</a>
+ */
+@Documented
+@DefaultBoolean(false)
+@Parameter(Boolean.class)
+@Qualifier
+@Target({ElementType.METHOD, ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Intercept {
```
mdekstrand: Maybe call this `IncludeIntercept`?

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
KimuraTian: Yes, I saw `ebeMultiply` in `util.math.RowView`. We can optimize it.

lenskit-slim/src/main/java/org/lenskit/slim/Intercept.java
KimuraTian: Do we need to remove it as so far it doesn't support updating with intercept?

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
KimuraTian: We move it in this feature branch?

lenskit-slim/src/main/java/org/lenskit/slim/MinCoRatedItems.java
KimuraTian: Yep.

lenskit-slim/src/main/java/org/lenskit/slim/LinearRegressionHelper.java
mdekstrand: Yes, please.

lenskit-core/src/main/java/org/lenskit/util/math/Vectors.java
```diff
@@ -326,6 +328,93 @@ public static RealVector matrixRow(RealMatrix mat, int row) {
...
+        return mapT;
+    }
+
+    /**
+     * Compute the addition of two vectors
+     * @param a The first vector
+     * @param b The second vector
+     * @return A new vector of the addition of two vectors {@code a} + {@code b}
+     */
+    public static Long2DoubleMap combineAndLimit(Long2DoubleMap a, Long2DoubleMap b) {
```
mdekstrand: I think this one should be called `add`, and the old `combine` should be `combineAndLimit`.

KimuraTian: Yes, they were renamed.

KimuraTian:
I am finishing up HPF implementation. Hopefully, we can get it for your review within this week.

lenskit-slim/src/main/java/org/lenskit/slim/SLIMBuildContextProvider.java
```diff
@@ -0,0 +1,301 @@
...
+        }
+    }
+
+    private Long2ObjectMap<Long2DoubleAccumulator> makeAccumulators(LongSet items) {
+        Long2ObjectMap<Long2DoubleAccumulator> rows = new Long2ObjectOpenHashMap<>(items.size());
+        LongIterator iter = items.iterator();
+        while (iter.hasNext()) {
+            long item = iter.nextLong();
+            Long2DoubleAccumulator accum;
+            if (modelSize == 0) {
```
mdekstrand: Is this model size truncation something that Karypis et al did in their SLIM paper? Or did they just use the lasso regression?

lenskit-slim/src/main/java/org/lenskit/slim/SLIMScorer.java
```diff
@@ -0,0 +1,77 @@
...
+import javax.inject.Inject;
+import java.util.*;
+
+/**
+ * SLIM scorer
+ * Set itemScorer to this scorer in recommender configuration
+ *
+ * @author <a href="http://www.grouplens.org">GroupLens Research</a>
+ */
+public class SLIMScorer extends AbstractItemScorer {
```
mdekstrand: This should be called `SLIMItemScorer`

lenskit-slim/src/main/java/org/lenskit/slim/SLIMScorer.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Score items for a user.
+     * @param user The user ID.
+     * @param items The score vector.
+     */
+    @Override
+    public ResultMap scoreWithDetails(long user, @Nonnull Collection<Long> items) {
+        logger.debug("scoring {} items for user {} with details", items.size(), user);
+        Long2DoubleMap ratings = Long2DoubleSortedArrayMap.create(rvDAO.userRatingVector(user));
```
mdekstrand: We need to normalize the ratings and then denormalize the predictions.

lenskit-slim/src/main/java/org/lenskit/slim/SLIMScoringStrategy.java
mdekstrand: Then let's consider dropping it, if it is easy enough to write the tests with the other method.

lenskit-slim/src/main/java/org/lenskit/slim/SLIMBuildContextProvider.java
```diff
@@ -0,0 +1,301 @@
...
+
+    /**
+     * Build item rating matrix
+     * @param itemVectors mapping from item ids to (userId: rating) maps to be filled
+     * @param userItems mapping from user ids to sets of user rated items (to be filled)
+     */
+    private void buildItemRatings(Long2ObjectMap<Long2DoubleSortedMap> itemVectors,
+                                  Long2ObjectMap<LongSortedSet> userItems) {
+        Long2ObjectMap<LongSet> userItemsTemp = new Long2ObjectOpenHashMap<>();
+        try (ObjectStream<IdBox<List<Rating>>> stream = dao.query(Rating.class)
```
mdekstrand: We really need to use user-vector normalization for consistency. Also, we should use the `RatingVectorPDAO` instead of a data access object.

That means we will be processing ratings user-by-user instead of item-by-item. Once the final context is built, there will be no difference, but the loop here will need to change.

KimuraTian: Yes, we can fix it

lenskit-slim/src/main/java/org/lenskit/slim/SLIMBuildContextProvider.java
KimuraTian: In that paper, there are two types of SLIM. One is using all the items except the label item itself as neighborhood items to train the regression model, the other one is using cosine similarity to find neighborhood items and then using those items to train the regression.

KimuraTian: The paper calls it fsSLIM. So the model size is the number of neighbors. The paper used 100 and 150

mdekstrand: Ok, so that's what you were doing with the `ItemItemModel` in the old code. I did not understand this.

It looks like, to fix that, we now have a copy of item-item's model-build here.

I propose: we have two different context providers (with appropriate code sharing), one for fsSLIM and one for the other one. The fsSLIM can use an item-item model, as that will have the cosine similarities pre-computed.

KimuraTian: Yes, we now have a copy of item-item's model build here.

When `SLIMModelSize` is set to 0, it's SLIM otherwise fsSLIM

KimuraTian: Actually, `ItemItemModel` itself has `ModelSize`. By setting it to zero (or not), we get SLIM (or fsSLIM).

mdekstrand: I think it would be better to have 2 different implementations in this case, rather than depending on a parameter value. There are several ways we can design this; since the only difference between SLIM and fsSLIM seems to be how the 'candidate neighbors' are selected, we can have a `CandidiateNeighborSelector` interface that abstracts the difference.

It's often good design, when we identify some piece of logic that has 2 configurable variants, to abstract it into an interface & implementations (the *strategy* design pattern), so we can reconfigure it. The fsSLIM implementation can then depend on knn's `ItemItemModel` in order to identify candidate neighbors.

KimuraTian: Thank you so much! I am trying to do so. :)

lenskit-slim/src/main/java/org/lenskit/slim/SLIMScorer.java
KimuraTian: Normalizer added.

lenskit-slim/src/main/java/org/lenskit/slim/SLIMBuildContextProvider.java
KimuraTian: Added a provider `FsSLIMBuildContextProvider` which depends on knn's `ItemItemModel`
The `SLIMBuildContextProvider` now get all the items as neighbors.

lenskit-slim/src/main/java/org/lenskit/slim/SLIMScorer.java
KimuraTian: It was renamed.

KimuraTian:
So far the predictions can be within the range from 0 to 5, and seems reasonable using simulated ratings. It needs further to be tested by real data.

PR #1000, opened 2017-03-16
Ignore null values in JSON entity format

mdekstrand:


mdekstrand:
@ANUYAH1 This will fix the problem that you encountered.

Issue #1003, opened 2017-03-17
Support Bellogin's evaluation methods


mdekstrand:
The crossfolder and train-test evaluator need to support Bellogin's various popularity-nullification eval techniques, from [his dissertation](http://ir.ii.uam.es/~alejandro/thesis/thesis-bellogin.pdf).

- [x] U1R/UAR: support crossfold splitting by item
- [x] 1R: support evaluating separately for each test item, not just each user

PR #1004, opened 2017-03-17
Extend top-N evaluation methods

mdekstrand:
This addresses #1003, making sure that we support Bellogin's various evaluation methods.


lenskit-eval/src/main/java/org/lenskit/eval/traintest/recommend/RecommendEvalTask.java
```diff
@@ -258,6 +303,10 @@ public void addMetric(TopNMetric<?> metric) {
 
     @Override
     public List<String> getUserColumns() {
+        if (separateItems) {
+            Collections.emptyList();
```
kluver: Just so I'm clear, we do this because we use a different file for the seperated (1r) evaluations. right?

mdekstrand: Yes.

mdekstrand:
@kluver Yes, half of this is for crossfolding by item (to implement Bellogin's UAR/U1R methods).

kluver:
I will review UAR/U1R. I don't remember Bellogin talking about item based crossfolds instead of user based crossfolds. It's been a while since I last reviewed Bellogin.

mdekstrand:
@kluver He doesn't call them item-based crossfolds, but they're what the 'U' family are. Uniform # of test ratings per item, so that popular items don't bias results just be being the right answer more often.

kluver:
Found that, section in thesis between classes. Will review tonight.

kluver:
Having trouble 100% decoding some of the math relations proposed and parsing them into operations. but I think, if nothing else, a per-item crossfold with a carefully  specified retain number can satisfy the sprit of the uniform-* methods. For a item-crossfold with a given retain, if an item has fewer than the retain number ratings, will we retain all ratings, or no ratings and exclude the item from evaluation? I think bellogin would have us not evaluate for items with fewer than the selected number of items, and I thought lenskit would put all items into the test set.

mdekstrand:
LensKit will put them in the test set, but that can be fixed with pre- or post-filtering. Or we can add an option.

Issue #999, opened 2017-03-14
ItemRecommender documentation is vague on some details.


kluver:
The `ItemRecommender` interface is vague. In particular:

- negative numbers are documented to indicate "no particular size requested" however, it is convention (which I want to rely on) that -1 should indicate "rank all items"
- It is not made clear that implementations are allowed to return fewer than the requested n items (such as if the item recommender enforces a constraint and cannot or will not return more items)

mdekstrand:
> negative numbers are documented to indicate "no particular size requested" however, it is convention (which I want to rely on) that -1 should indicate "rank all items"

Hmm. I sympathize. However, I am very hesitant to increase the guarantees implementations are required to maintain.

> It is not made clear that implementations are allowed to return fewer than the requested n items (such as if the item recommender enforces a constraint and cannot or will not return more items)

Definitely need to clear that up.

kluver:
Quoting you on PR997 on the subject of -1 == all:
> It does. Actually, any negative value is supposed to be all.
Currently there is no "correct" way to ask for an absolute ranking that I can think of. I consider this a problem.

mdekstrand:
> Currently there is no "correct" way to ask for an absolute ranking that I can think of. I consider this a problem.

I do not automatically disagree.

In conversation with Qian, we've discussed the relationship of *retrieval*, *ranking*, and *recommendation*.

What would you think about specifying that -1 means 'recommend as many as possible'? Leaves the door open for some hypothetical recommender with a strong belief that 10 items are all you need, but much more strongly indicates that any recommender *without* such a belief should generate as long a list as it can.

kluver:
I like this resolution. I agree that not all recommenders can produce a complete list (I just built one that will provide a truncated list due to constraints)

So the changes:

- [ ] Document that recommenders are not guaranteed to return n items
- [ ] Document that negative means "as many as possible" which often implies "all", but not always.

PR #1005, opened 2017-03-18
Update item recommender docs

mdekstrand:
This updates the item recommender docs for #999.


mdekstrand:
That one build failure is ephemeral and fine.

Issue #1001, opened 2017-03-17
Support query/runtime data in train-test evaluation


mdekstrand:
The train-test evaluator needs to support 'query' or 'runtime' data, in addition to 'train' data, so that we can simulate having additional data available at runtime that is not the test data.

This is half of #561.

PR #1006, opened 2017-03-18
Add runtime data support

mdekstrand:
This PR adds runtime data (#1001) to train-test, and removes the old `@QueryData` mechanism that was weird and no longer worked.


mdekstrand:
This is very memory-inefficient, but it will get the job done.

We should support somehow making the runtime and train data share storage to the extent possible.

PR #1013, opened 2017-04-12
Improve entity code

mdekstrand:
This works on improving the entity code to support more efficient entity objects and data sources.

* Attribute sets decrease overhead of storing multiple copies of entities' attribute names
* `AbstractBeanEntity` and `AbstractBeanEntityBuilder` make it easier to implement entity view classes
* `Text` allows us to more efficiently store text data in memory

### DIFF
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractBeanEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractBeanEntity.java
@@ -1,3 +1,23 @@
+/*
+ * LensKit, an open source recommender systems toolkit.
+ * Copyright 2010-2016 LensKit Contributors.  See CONTRIBUTORS.md.
+ * Work on LensKit has been funded by the National Science Foundation under
+ * grants IIS 05-34939, 08-08692, 08-12148, and 10-17697.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
 package org.lenskit.data.entities;
 
 import com.google.common.collect.ImmutableList;
--- a/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractBeanEntity.java
+++ b/lenskit-core/src/main/java/org/lenskit/data/entities/AbstractBeanEntity.java
@@ -1,13 +1,11 @@
 package org.lenskit.data.entities;
 
-import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.reflect.TypeToken;
 import org.apache.commons.lang3.tuple.Pair;
 
 import javax.annotation.Nullable;
-import java.lang.invoke.MethodHandle;
-import java.lang.invoke.MethodHandles;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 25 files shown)


lenskit-core/src/main/java/org/lenskit/data/entities/AbstractBeanEntityBuilder.java
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * Base class for entity builders using bean-style setters.  An entity builder should extend this class, then annotate
+ * setter methods with {@link EntityAttributeSetter} (and clearers with {@link EntityAttributeClearer}).  The standard
+ * methods will then be implemented in terms of these.
+ */
+public class AbstractBeanEntityBuilder extends EntityBuilder {
+    private static final ConcurrentHashMap<Class<? extends AbstractBeanEntityBuilder>, AttrEntry[]> cache =
+            new ConcurrentHashMap<>();
+
+    private final AttrEntry[] attributeRecords;
```
kluver: Its not clear to me why this isn't a map from names to pairs of setter and getter.

lenskit-core/src/main/java/org/lenskit/data/entities/TypedName.java
```diff
@@ -65,7 +65,7 @@ private TypedName(String n, TypeToken<T> t) {
 
     /**
      * Get the underlying name.
-     * @return The name.
+     * @return The name. This is guaranteed to be an interned string (see {@link String#intern()}).
```
kluver: Yay! I learned about these the other day and was thinking we should do this.

lenskit-core/src/main/java/org/lenskit/util/Text.java
```diff
@@ -0,0 +1,84 @@
...
+import org.joda.convert.ToString;
+
+import java.io.Serializable;
+import java.text.Normalizer;
+import java.util.Arrays;
+
+/**
+ * UTF-8 encoded text string.
+ */
+public final class Text implements Serializable {
```
kluver: I find it hard to believe that we had to implement utf-8 strings for ourselves.

lenskit-core/src/main/java/org/lenskit/data/entities/AbstractBeanEntity.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Abstract entity implementation that uses bean methods.
+ */
+public abstract class AbstractBeanEntity extends AbstractEntity {
+    private static final ConcurrentHashMap<Class<? extends AbstractBeanEntity>, Pair<AttributeSet,ImmutableList<Method>>> cache =
+            new ConcurrentHashMap<>();
+
+    protected final AttributeSet attributes;
```
mdekstrand: Crap, did I fail to put comments in as to *why* I did this?

There are two purposes for this.

1. Speed up lookup. I micro-benchmarked hash tables, identity hash tables, and a quick linear search through an array with `==`, and the linear search won for small numbers of attributes, on the order of the number of attributes a typical entity has.
2. Decrease memory use. Since attribute sets are constant and interned, each entity with the same set of attributes shares the storage for the attribute names. This isn't as big a deal for the bean entities, since we're caching those, but this lets me use identical attribute management logic between this and the basic entity (and soon the packed entity storage that I am currently working on).

lenskit-core/src/main/java/org/lenskit/data/entities/AbstractBeanEntity.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+
+    @Override
+    public int getInteger(TypedName<Integer> name) {
+        int idx = attributes.lookup(name);
+        if (idx >= 0) {
+            Method mh = methods.get(idx);
+            try {
+                return (int) mh.invoke(this);
```
mdekstrand: Yes, but that's cheap on modern JVMs. I have tried to use the new `MethodHandle` - which might let us not box/unbox, theoretically - but it was slower.

Non-boxing/unboxing solutions will require runtime code generation, I think.

lenskit-core/src/main/java/org/lenskit/data/entities/AbstractBeanEntityBuilder.java
mdekstrand: The speed of linear search again. Should add comments?

lenskit-core/src/main/java/org/lenskit/data/entities/TypedName.java
mdekstrand: And this is what makes linear search fast :)

lenskit-core/src/main/java/org/lenskit/util/Text.java
mdekstrand: Yes, we did. I found 2 other implementations of UTF-8 strings for Java: one lives in Hadoop, and the other lives in Apache Avro, which is a full serialization framework that we don't otherwise need.

So I built it, and was none too pleased about the need to do so.

But it will significantly reduce our memory use when storing a bunch of strings in entities in memory, because Java uses UTF-16.

mdekstrand: Also, fun story: this doesn't implement `CharSequence` - though I have the code on-hand to do that - because of JodaOrg/joda-convert#15.

mdekstrand:
@kluver Thanks for looking! The user-facing aspect of this is how much simpler `Rating` and `RatingBuilder` are now.

lenskit-core/src/main/java/org/lenskit/data/entities/AbstractBeanEntity.java
kluver: I would comment this becuase you did micro-benchmarks.

lenskit-core/src/main/java/org/lenskit/data/entities/AbstractBeanEntity.java
kluver: If this does boxing anyways why do we have special case methods for Integer etc? Shouldn't those just fall under the generic get<T> method (I remember seeing such a method, not opining up code to find its name)

mdekstrand: That is... a good question. A very good question.

For these abstract bean entities, it'll box (I think). For the basic entity, things are boxed. I would like to make the abstract bean entities not box, but that *looks* like it will require codegen.

For  the packed entity storage I am currently building, I think I'll be able to do it without boxing.

Not sure if it's worth it.

Issue #969, opened 2016-11-04
Implement packed entity DAO


mdekstrand:
To decrease memory usage, we should implement a packed entity DAO that stores entities in parallel arrays or buffers, possibly with flyweights for accessing them.

This will be like the old packed entity rating DAO, but transparent and only in-memory.

If we use buffers, then it can play nice with swap.

mdekstrand:
This may make #948 easier to implement.

mdekstrand:
Long-term, I would very much like to use Apache Arrow for this.

PR #1015, opened 2017-04-14
Packed entities

mdekstrand:
This will work on #969, packed entity support in the DAO. It also introduces code generation and uses it to reduce the overhead.

Microbenchmarks indicate that low-level entity operations are slower - quite a bit slower - but overall experiment performance seems to be unaffected in at least one example experiment.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -133,7 +133,6 @@ subprojects {
     }
 
     // don't package until testing
-    jar.mustRunAfter 'test'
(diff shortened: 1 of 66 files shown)


mdekstrand:
Microbenchmarks show that really simple processing of a rating data file is notably slower with this code. However, a real experiment sped up by almost 10%, which I am more inclined to believe.

Issue #1016, opened 2017-04-18
Add setting to restrict parallel evaluations


mdekstrand:
After #1011, the evaluator thread pool is used both to run multiple evaluations in parallel, and to parallelize a single evaluation.

For memory use, we should add an option to limit the number of concurrent eval jobs independent of the thread count, so that we can parallelize evaluation (huge win) without the memory load of multiple models in memory simultaneously.

PR #1017, opened 2017-04-18
Add more evaluation control

mdekstrand:
This allows us to restrict the number of simultaneous evaluation jobs independently from the thread count.

The default is to use all threads.

**Note for reviewer:** this also changes the actual default thread count to match the documented default of all available processors.

This works on #1016.

### DIFF
--- a/lenskit-eval/src/main/java/org/lenskit/eval/traintest/ExperimentJob.java
+++ b/lenskit-eval/src/main/java/org/lenskit/eval/traintest/ExperimentJob.java
@@ -22,6 +22,7 @@ package org.lenskit.eval.traintest;
 
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.Lists;
+import net.jcip.annotations.ThreadSafe;
 import org.grouplens.grapht.Component;
 import org.grouplens.grapht.Dependency;
 import org.grouplens.grapht.InjectionException;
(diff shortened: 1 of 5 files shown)


mdekstrand:
Requesting review from @kluver on the change in default value for the thread count.

lenskit-eval/src/main/java/org/lenskit/eval/traintest/ExperimentJob.java
```diff
@@ -22,6 +22,7 @@
```
kluver: Uggh the diff on this class is miserable. Given only a cursory review this appears to mosty a restructure of already existing code plus added blocking on the semaphore.

mdekstrand: That's exactly what it is. Separated into separate commits, but that doesn't help in the PR diff. Adding the semaphore alone would have made an already-terrible method even *more* indented.

mdekstrand:
Glad you like the docs. Since you didn't say they aren't, I assume you find the defaults also sensible.

kluver:
I hardly even noticed the defaults. "gotta go fast" is a great default setting as far as I'm concerned. and the impact of reducing it (better memory performance, possibly not much worse speed) seems well documented.

PR #1027, opened 2017-05-31
[WIP] add BPR-MF to lenskit.

kluver:
Work in progress pull request on BPR-MF being added to lenskit. I like feedback on the design and a few choices I wasn't particularly sure how I wanted to deal with. I also havn't tested this version front-to-back and I'm curious how you would advise testing it (other than then building an algorithm over movielens somewhere and seeing if it looks like it should work, since that doesn't seem particularly robust)


mdekstrand:
So, front-to-back tests - yep, running over MovieLens is what we do. There are typically 2 versions of that:

1. A 'builld-serialize' test that makes sure we can train a model, save it, and re-load it.
2. An 'accuracy' test that checks if a metric is within tolerance of an expected value.

Other than that... individual component unit tests are helpful but often tricky.

kluver:
Hmm... I don't think I can unit test any more of the components than I already am. I have three untested class, the algorithm itself, a random rating pair generator which (as its random) I can't think of a great way to test, and a POJO object so simple it could be a `long[3]` without much loss of generality.

I'll look into a build-serialize test. That said I don't know that it is necessary, I don't implement a new type of model, I just build an MFModel. (I was going to say like the weighted MF model builder, but that doesn't seem to be a thing any more). 

As for an accuracy test (accuracy would be the wrong word here BTW) I don't know that we have third party comparison for metric values of this algorithm on MovieLens. (I also don't know if we have the most appropriate metric ready for comparison either). I could run it and say "it should be around what it is right now", but I don't know what that proves.

I think the "accuracy test" is what needs the most thought/creativity here. I'll look into that one.

mdekstrand:
> As for an accuracy test (accuracy would be the wrong word here BTW) I don't know that we have third party comparison for metric values of this algorithm on MovieLens. (I also don't know if we have the most appropriate metric ready for comparison either). I could run it and say "it should be around what it is right now", but I don't know what that proves.

It lets us catch changes which induce significant change in accuracy. It doesn't on its own prove anything, but it helps us future-proof by flagging code changes that make a big difference for further scrutiny.

mdekstrand:
The build-serialize test will do 2 things:

1. End-to-end smoke test to make sure that training terminates & doesn't crash
2. Make sure the dependency structure allows models to be serialized

lenskit-core/src/main/java/org/lenskit/util/math/RollingWindowMeanAccumulator.java
```diff
@@ -0,0 +1,114 @@
...
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.util.math;
+
+import java.util.Arrays;
+
+/**
+ * Accumulates the mean of the most recent accumulated values. Meaning that its reported mean will only reflect the N
+ * most recent values. This is accomplished using a fixed size array and a running total. Adding and getting the average
+ * are take O(1).
```
mdekstrand: need a little grammar editing here - 'are take O(1)'

lenskit-core/src/main/java/org/lenskit/util/math/RollingWindowMeanAccumulator.java
```diff
@@ -0,0 +1,114 @@
...
+ */
+public class RollingWindowMeanAccumulator {
+    private static final int DEFAULT_SIZE = 100;
+
+    private final double[] data;
+    // invariant - sum of data.
+    private double sum;
+    // invariant - if position is negative then the next index is data.length + position (and we are filling the data
+    // array for the first time) otherwise position is the next index (meaning the index of the oldest value).
+    private int position;
```
mdekstrand: Interesting design here - not quite how I would have done it, but that probably doesn't matter.

I would use two variables, `next` and `size`, with the update rule `next = (next + 1) % size`.

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BPRMFModelProvider.java
```diff
@@ -0,0 +1,175 @@
...
+        // and is therefore a reasonable "max_error".
+        RollingWindowMeanAccumulator optAccum = new RollingWindowMeanAccumulator(1000, -30);
+
+        int userCount = dao.getEntityIds(CommonTypes.USER).size();
+        RealMatrix userFeatures = MatrixUtils.createRealMatrix(userCount, featureCount);
+        initFeatures(userFeatures);
+
+        int itemCount = dao.getEntityIds(CommonTypes.ITEM).size();
+        RealMatrix itemFeatures = MatrixUtils.createRealMatrix(itemCount, featureCount);
+        initFeatures(userFeatures);
```
mdekstrand: Needs to be `itemFeatures`?

mdekstrand: Needs to be `itemFeatures`?

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BPRMFModelProvider.java
```diff
@@ -0,0 +1,175 @@
...
+        TrainingLoopController controller = stoppingCondition.newLoop();
+
+        //REVIEW: because of the nature of training samples (and the point that the BPR paper makes that training
+        // by-item or by-user are not optimal) one "iteration" here will be one training update. This leads to _really_
+        // big iteration counts, which can actually overflow ints!. one suggestion would be to allow the iteration count
+        // controller to accept longs, but I don't know if this will introduce backwards compatibility issues (I imagine
+        // this depends on the robustness of our type conversion in the configuration.
+        while(controller.keepTraining(-optAccum.getMean())) {
+            TrainingItemPair pair = pairGenerator.nextPair();
+            // Note: bad code style variable names are generally to match BPR paper and enable easier implementation
```
mdekstrand: I don't have a problem with it.

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BPRMFModelProvider.java
```diff
@@ -0,0 +1,175 @@
...
+        logger.info("Building MPR-MF with {} features for {} users and {} items",
+                featureCount, userCount, itemCount);
+
+        TrainingLoopController controller = stoppingCondition.newLoop();
+
+        //REVIEW: because of the nature of training samples (and the point that the BPR paper makes that training
+        // by-item or by-user are not optimal) one "iteration" here will be one training update. This leads to _really_
+        // big iteration counts, which can actually overflow ints!. one suggestion would be to allow the iteration count
+        // controller to accept longs, but I don't know if this will introduce backwards compatibility issues (I imagine
+        // this depends on the robustness of our type conversion in the configuration.
```
mdekstrand: Good question. We should test. I suspect Groovy will be obnoxious here.

Could we scale the iteration counts? use pseudo-counts that are `ct / nratings`? Would need to change interfaces.

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BPRMFModelProvider.java
```diff
@@ -0,0 +1,175 @@
...
+            long iid = pair.g;
+            int i = itemIndex.internId(iid);
+            long jid = pair.l;
+            int j = itemIndex.internId(jid);
+            long uid = pair.u;
+            int u = userIndex.internId(uid);
+
+            RealVector w_u = userFeatures.getRowVector(u);
+            RealVector h_i = itemFeatures.getRowVector(i);
+            RealVector h_j = itemFeatures.getRowVector(j);
```
mdekstrand: [`Vectors.matrixRow`](http://lenskit.org/master/apidocs/org/lenskit/util/math/Vectors.html#matrixRow-org.apache.commons.math3.linear.RealMatrix-int-) will return a flyweight backed by the matrix; `getRowVector` will copy, I believe.

mdekstrand: fun trick: that flyweight will probably be stack-allocated after the inliner does its magic.

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BPRMFModelProvider.java
```diff
@@ -0,0 +1,175 @@
...
+            h_i_update.combineToSelf(1, bprTerm, w_u);
+
+            // h_j update
+            RealVector h_j_update = h_j.mapMultiply(-regularization);
+            h_j_update.combineToSelf(1, -bprTerm, w_u);
+
+            // perform updates
+            w_u.combineToSelf(1, learningRate, w_u_update);
+            h_i.combineToSelf(1, learningRate, h_i_update);
+            h_j.combineToSelf(1, learningRate, h_j_update);
```
mdekstrand: This won't actually update without calls to the `setRowVector` method on the matrix..

lenskit-svd/src/main/java/org/lenskit/mf/BPR/TrainingItemPair.java
```diff
@@ -0,0 +1,38 @@
...
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.lenskit.mf.BPR;
+
+/**
+ * class to represent a training pair in pairwise learning to rank items.
+ * By convention the pair represents the knowledge that g>l (according to user u)
+ */
+public class TrainingItemPair {
+    // REVIEW: if I make these ints (rows in the matrix as per the snapshot) the code would theoretically be faster.
```
mdekstrand: I like the generality of longs. Let's only make them `int`s if we measure a bottleneck.

lenskit-svd/src/main/java/org/lenskit/mf/BPR/RandomRatingPairGenerator.java
```diff
@@ -0,0 +1,83 @@
...
+import it.unimi.dsi.fastutil.longs.LongList;
+import org.lenskit.data.ratings.RatingMatrix;
+
+import javax.inject.Inject;
+import java.util.Random;
+
+/**
+ * Randomly generates training pairs by selecting a random user and two random rated items.
+ * If the two random rated items have different ratings, than those items are returned as a training pair.
+ * Otherwise, the algorithm will loop until it finds a training pair.
```
mdekstrand: maybe explicitly mention that this uses rejection sampling?

lenskit-svd/src/main/java/org/lenskit/mf/BPR/RandomRatingPairGenerator.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return out;
+    }
+
+    private TrainingItemPair tryGenTrainingPair() {
+        // randomly generate training pairs, return
+
+        long userId = userIds.getLong(rand.nextInt(userIds.size()));
+        Long2DoubleMap ratings = snapshot.getUserRatingVector(userId);
+        long[] ratedItems = ratings.keySet().toLongArray();
```
mdekstrand: eek this feels slow. no better way to do it?

lenskit-svd/src/main/java/org/lenskit/mf/BPR/RandomRatingPairGenerator.java
```diff
@@ -0,0 +1,83 @@
...
+
+import javax.inject.Inject;
+import java.util.Random;
+
+/**
+ * Randomly generates training pairs by selecting a random user and two random rated items.
+ * If the two random rated items have different ratings, than those items are returned as a training pair.
+ * Otherwise, the algorithm will loop until it finds a training pair.
+ */
+public class RandomRatingPairGenerator implements TrainingPairGenerator{
```
mdekstrand: way to test:

- generate 100 pairs
- for each, verify that user & item IDs are valid, and that r_ug > r_ul.

lenskit-core/src/main/java/org/lenskit/util/math/RollingWindowMeanAccumulator.java
kluver: It came to me in a flash of "this would work"

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BPRMFModelProvider.java
kluver: Oh crap! yah, I expect it does!

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BPRMFModelProvider.java
kluver: I was thinking about this, if we want to use any two count system I think the lower level of counts should go into the training pair generator. That gives the flexibility to do exhaustive all pairs generators in a clean way, I'm just not sure I like the idea of a two count system.

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BPRMFModelProvider.java
kluver: I'll look into this. I honestly don't understand the massive variety of matrix methods, I think I only just got my head around the old library when we switched to a new one.

lenskit-svd/src/main/java/org/lenskit/mf/BPR/TrainingItemPair.java
kluver: Agreed. that was my intuition as well.

lenskit-svd/src/main/java/org/lenskit/mf/BPR/RandomRatingPairGenerator.java
kluver: Its something I suppose.

lenskit-svd/src/main/java/org/lenskit/mf/BPR/RandomRatingPairGenerator.java
kluver: I don't know, the dao interfaces confuse me. I'll poke and see if I can think of better. If we want to pay the memory costs I could at least cache, memoize, or build a look-up map.

lenskit-core/src/main/java/org/lenskit/util/math/RollingWindowMeanAccumulator.java
kluver: fixed

kluver:
I've added a serialize test (which was quite useful in finding issues) and an accuracy test (which I'm not convinced I belive). Right now the accuracy test uses the ndcg predict metric, since I want an overall ranking metric. Also, in its current configuration it doubles test time.

Question: Is it worth building out a AUC metric for this PR?
Question: is it worth independently tuning the number of training iterations

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BatchSize.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * The number of iterations to use in an iterative update.
+ */
+@Documented
+@Parameter(Integer.class)
+@Qualifier
+@Target({ElementType.METHOD, ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface BatchSize {
```
mdekstrand: default values?

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BPRMFModelProvider.java
mdekstrand: I think that the answer here is that the training loop controller is a failed abstraction. We also couldn't make it work for poisson factorization.

mdekstrand: #1046 has a draft version of removing stopping conditions.

mdekstrand: We need to stop using training loop controllers, in favor of a more direct approach, before we can merge.

lenskit-svd/src/main/java/org/lenskit/mf/BPR/BPRMFModelProvider.java
```diff
@@ -0,0 +1,183 @@
...
+        for(long iid : dao.getEntityIds(CommonTypes.ITEM)) {
+            itemIndex.internId(iid);
+        }
+        RealMatrix itemFeatures = MatrixUtils.createRealMatrix(itemCount, featureCount);
+        initFeatures(itemFeatures);
+
+        logger.debug("Learning rate is {}", learningRate);
+        logger.debug("Regularization term is {}", regularization);
+
+        logger.info("Building MPR-MF with {} features for {} users and {} items",
```
mdekstrand: BPR-MF

PR #1052, opened 2018-04-27
Add BPR to LensKit

mdekstrand:
This supersedes #1027, adding BPR to LensKit with the current code base.

Still probably bugs to fix!

### DIFF
--- a/lenskit-eval/src/main/java/org/lenskit/eval/traintest/recommend/TopNAUCMetric.java
+++ b/lenskit-eval/src/main/java/org/lenskit/eval/traintest/recommend/TopNAUCMetric.java
@@ -44,6 +44,9 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.List;
+import java.util.BitSet;
+import java.util.function.LongPredicate;
+import java.util.stream.IntStream;
 
 /**
  * Compute the area under the ROC curve.
--- a/lenskit-eval/src/test/java/org/lenskit/eval/traintest/recommend/TopNAUCMetricTest.java
+++ b/lenskit-eval/src/test/java/org/lenskit/eval/traintest/recommend/TopNAUCMetricTest.java
@@ -1,0 +1,59 @@
+/*
+ * LensKit, an open-source toolkit for recommender systems.
+ * Copyright 2014-2017 LensKit contributors (see CONTRIBUTORS.md)
+ * Copyright 2010-2014 Regents of the University of Minnesota
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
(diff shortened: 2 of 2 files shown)


lenskit-core/src/main/java/org/lenskit/util/XoRoShiRo128PlusRandomGenerator.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.commons.math3.random.AbstractRandomGenerator;
+
+import javax.inject.Inject;
+
+/**
+ * XoRoShiRo128+ random number generator. This is implemented from the public-domain sources by David Blackman
+ * and Sebastiano Vigna at <http://xoroshiro.di.unimi.it/xoroshiro128plus.c> and the notes at
+ * <http://xoroshiro.di.unimi.it/>.
+ */
+public class XoRoShiRo128PlusRandomGenerator extends AbstractRandomGenerator {
```
kluver: Convince me we need this.

lenskit-core/src/test/java/org/lenskit/util/XoRoShiRo128PlusRandomGeneratorTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Test
+    public void testBooleans() {
+        int ntrue = 0;
+        for (int i = 0; i < 10000; i++) {
+            if (rng.nextBoolean()) {
+                ntrue += 1;
+            }
+        }
+        assertThat(ntrue, greaterThan(2500));
```
kluver: Can we get tighter confidence bound tests here. Not that I like more false-accepts, its just this feels pretty generous and might miss systematic errors of a certain scale (I can't commit to those statistics right now ...)

lenskit-core/src/main/java/org/lenskit/util/XoRoShiRo128PlusRandomGenerator.java
mdekstrand: I have had some nagging concerns regarding the suitability of standard (i.e. bad) RNGs for the kinds of purposes we put them to in LensKit, especially with high-sample-count operations like BPR training. No concrete evidence of a problem, but Java Random is a basic `rand`, and far behind the state of the art; XoRoShiRo+ is also supposed to be a lot faster.

I wanted to rule out low-quality training pair sampling as a cause of my observed BPR woes. I knew a utility package by the author of `fastutil` had another RNG that was claimed to be much better; unfortunately, it is license-incompatible, but Vigna has a public domain C implementation of it, which I ported to Java here.

Swapping out the RNG made no difference in the problems I was seeing. Possibly silly thing to try in the first place. However, since I wrote the code, keeping it around lets us test RNG impact on LensKit runs and evaluation. It is disabled by default - need to rebind `RandomGenerator` to make use of it.

lenskit-core/src/test/java/org/lenskit/util/XoRoShiRo128PlusRandomGeneratorTest.java
mdekstrand: It'd be nice, I haven't done any meaningful statistics to create these bounds.

My goal was 'easy way to make sure that it is actually generating different bits, not just the same bit all the time'.

